PR #107, opened 2020-08-10
GRIPPS: Fix NullPointerException when -output_vcf is a file name in CWD

vladsaveliev:
GRIPPS's failing when the input VCF is a relative file path in CWD, e.g. `-output_vcf gridss.somatic.vcf.gz`, because `File("gridss.somatic.vcf.gz").parentFile` is null.

Adding an extra `absoluteFile` fixes the issue.

Vlad


jonbaber:
Thanks Vlad. There is one other outstanding bug from Dario I need to fix today then I should be able to release a new version.

PR #111, opened 2020-08-14
GRIPSS: support tumor-only?

vladsaveliev:
I know your pipeline is designed for tumor/normal paired variant calling; but sometimes we have tumor-only cases and I still find GRIDSS useful. So I slightly modified GRIPSS to work for tumor-only VCFs. 

Feel free to reject the PR if you didn't plan to support this case or if you wanted to implement it better (I'm not even sure if my way is correct!). Just wanted to bring this use case to you.

By the way, really appreciate the usage of Kotlin!

Vlad

### DIFF
--- a/gripss/src/main/kotlin/com/hartwig/hmftools/gripss/GripssApplication.kt
+++ b/gripss/src/main/kotlin/com/hartwig/hmftools/gripss/GripssApplication.kt
@@ -73,9 +73,12 @@ class GripssApplication(private val config: GripssConfig) : AutoCloseable, Runna
         logger.info("Config ${config.filterConfig}")
         val contigComparator = ContigComparator(dictionary)
 
-        val sampleNames =  inputHeader.genotypeSamples!!
-        val sampleOrdinals = sampleOrdinals(sampleNames);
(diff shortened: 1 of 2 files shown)


jonbaber:
Hi Vlad,

Thanks very much for the suggestion and the implementation. Glad you appreciate the Kotlin, this is my first project with it and really liking it so you will probably see more of it from me 👍 

Please don't let this discourage you from submitting pull requests in the future as it really does make my job easier, but I am going to reject this one only because I have slightly rearranged how we determine if we are in tumor-only mode. I just enforced that if you supply a sample name it still has to match what is in the file. I also added a couple of extra test cases. The main filtering code is still all from you!

I will have a new release for you in the next day or two.

Jon

vladsaveliev:
Fantastic, thanks Jon! Glad the idea made sense and tumor-only is going to be supported. And looking forward to more Kotlin!

Vlad

jonbaber:
No worries. Changes are now in master if you want an early preview.

jonbaber:
I have published the release: https://github.com/hartwigmedical/hmftools/releases/tag/gripss-v1.5

PR #159, opened 2021-03-03
DEV-1853 Create list of runcontexts from json when specified

pauldwolfe:
Adds a simple json deserializer which takes a local file and creates a list of run
contexts. Will be used when the run_dir argument is null and the run_json argument is
specified.

### DIFF
--- a/patient-db/src/main/java/com/hartwig/hmftools/patientdb/clinical/SampleLoader.java
+++ b/patient-db/src/main/java/com/hartwig/hmftools/patientdb/clinical/SampleLoader.java
@@ -55,21 +55,18 @@ public class SampleLoader {
     @NotNull
     private List<RunContext> extractRunContextsFromDirectoryOrJson(final @NotNull ClinicalAlgoConfig config, final String runsDirectory,
             final String runsJson) throws IOException {
-        List<RunContext> runContexts;
-        if (runsDirectory != null) {
+        if (runsDirectory != null && runsJson == null) {
             String pipelineVersionFile = config.pipelineVersionFile();
             if (pipelineVersionFile != null) {
                 LOGGER.info("   Loading sequence runs from {}", runsDirectory);
-                runContexts = loadRunContextsFromDirectory(runsDirectory, pipelineVersionFile);
+                return loadRunContextsFromDirectory(runsDirectory, pipelineVersionFile);
             } else {
                 throw new IllegalArgumentException("Cannot load from a runs directory if no pipeline version file given");
             }
-        } else if (runsJson != null) {
-            runContexts = loadRunContextsFromJson(runsJson);
-        } else {
-            throw new IllegalStateException("Either a runs directory or runs json must be specified");
+        } else if (runsJson != null && runsDirectory == null) {
+            return loadRunContextsFromJson(runsJson);
         }
-        return runContexts;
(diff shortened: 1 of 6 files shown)


patient-db/src/main/java/com/hartwig/hmftools/patientdb/clinical/SampleLoader.java
```diff
@@ -53,7 +53,35 @@ public SampleLoader(@NotNull final Lims lims) {
     }
 
     @NotNull
-    private static List<RunContext> loadRunContexts(@NotNull String runsDirectory, @NotNull String pipelineVersionFile) throws IOException {
+    private List<RunContext> extractRunContextsFromDirectoryOrJson(final @NotNull ClinicalAlgoConfig config, final String runsDirectory,
+            final String runsJson) throws IOException {
+        List<RunContext> runContexts;
+        if (runsDirectory != null) {
+            String pipelineVersionFile = config.pipelineVersionFile();
```
nedleitch, line 60: Given this is a new addition to an existing command line, should we consider also adding logging if both `runsDirectory` and `runsJson` are given (to tell the user that the JSON will be ignored essentially)?

patient-db/src/main/java/com/hartwig/hmftools/patientdb/clinical/context/ProductionRunContextFactory.java
```diff
@@ -20,4 +20,5 @@ public static RunContext fromRunDirectory(@NotNull String runDirectory, @NotNull
         }
         return runContextFromMetaData;
     }
+
```
nedleitch, line 23: Nit: this probably shouldn't have been added

patient-db/src/test/java/com/hartwig/hmftools/patientdb/clinical/readers/RunsJsonReaderTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void throwsIllegalArgumentOnEmptyFile() {
+        RunsJsonReader.extractRunContexts(testJson("empty.json"));
+    }
+
+    @Test
+    public void throwsIllegalArgumentOnUnparseableFile() {
+        RunsJsonReader.extractRunContexts(testJson("missing_fields.json"));
+    }
```
nedleitch, line 32: This looks to be missing the `expected =` attribute. And I don't see the JSON file?

nedleitch (changes requested):

patient-db/src/main/java/com/hartwig/hmftools/patientdb/clinical/SampleLoader.java
pauldwolfe: yes lets throw when neither are specified or both

patient-db/src/test/java/com/hartwig/hmftools/patientdb/clinical/readers/RunsJsonReaderTest.java
pauldwolfe: meant to remove this test

nedleitch (approved):

PR #206, opened 2021-09-08
Dev 2140

DFKoetsier:
Adjustments to allow patient-reporter to handle anonymized runs and normal runs in such a way that the final report is unchanged and unaffected by the difference. This adds two input arguments to the reporter. See [pilot_create_patient_report_for_run_anonymized](https://github.com/hartwigmedical/scripts/blob/master/hmftools/patientreporter/pilot_create_patient_report_for_run_anonymized) for how each of the sample id and barcode arguments needs to be provided. 

I think that the argument naming is currently a bit confusing, but I couldn't come up with anything better. Suggestions are welcome.


DFKoetsier:
My proposed changes hadn't taken (failed) samples without full runs into account. I have made adjustments and these samples should now also work. I have tested one such sample and found no issues.

DFKoetsier:
I thought of a better approach, so I'm closing this pull request.

PR #269, opened 2022-03-22
DEV-2623 Tumor only support in health checker

pauldwolfe:
Refactor the health checker to split the tumor/reference pairs and null checks on tumor, and
group together the per-sample arguments. Loosen restrictions on CLI args to allow for a
null ref as well as a null tumor.

ps. Germline only is already supported as "single sample", so we simply change some logging there.


health-checker/src/main/java/com/hartwig/hmftools/healthchecker/HealthChecksApplication.java
```diff
@@ -78,27 +72,22 @@ public static void main(String... args) throws ParseException, IOException {
...
+        String refSample = cmd.getOptionValue(REF_SAMPLE, null);
+        String refFlagstat = cmd.getOptionValue(REF_FLAGSTAT_FILE, null);
+        String refWgsMetricsFile = cmd.getOptionValue(REF_WGS_METRICS_FILE, null);
+        String tumorSample = cmd.getOptionValue(TUMOR_SAMPLE, null);
+        String tumorWgsMetricsFile = cmd.getOptionValue(TUMOR_WGS_METRICS_FILE, null);
+        String tumorFlagstat = cmd.getOptionValue(TUMOR_FLAGSTAT_FILE, null);
+
+        String purpleDir = cmd.getOptionValue(PURPLE_DIR, null);
+        new HealthChecksApplication(HealthCheckSampleConfiguration.of(tumorSample, tumorWgsMetricsFile, tumorFlagstat),
+                HealthCheckSampleConfiguration.of(refSample, refWgsMetricsFile, refWgsMetricsFile),
```
nedleitch, line 90: Think the last argument should be `refFlagstat`.

health-checker/src/test/java/com/hartwig/hmftools/healthchecker/HealthChecksApplicationTest.java
```diff
@@ -16,27 +17,30 @@
...
     @Test
     public void runHealthCheckerInSomaticMode() throws IOException {
-        HealthChecksApplication app = new HealthChecksApplication("reference",
-                "tumor",
-                BASE_DIR + "metrics/reference.wgsmetrics",
-                BASE_DIR + "metrics/tumor.wgsmetrics",
-                BASE_DIR + "flagstat/reference.flagstat",
-                BASE_DIR + "flagstat/tumor.flagstat",
+        HealthChecksApplication app = new HealthChecksApplication(HealthCheckSampleConfiguration.of("reference",
+                BASE_DIR + "metrics" + "/reference.wgsmetrics",
```
nedleitch, line 21: Unnecessary concatenation

nedleitch (changes requested):

PR #297, opened 2022-06-08
Dev 2739

DFKoetsier:
Hi Lieke and Korneel,

this is my refactor of the anonimisation of the patient-reporter.

I also encountered two spots in the code where the behaviour of the patient-reporter depended on whether the tumor sample name starts with "COLO" or not. This behaviour didn't seem ideal, so after discussing it with Lieke I have made some changes there too.

First, in ReportingDb.java the API-update JSON is always created, instead of only for some samples. Creating the JSON doesn't trigger any automatic API updates or anything, so this shouldn't cause any problems. 

Second, in SampleReportFactory.java the behaviour surrounding the "default" COLO cohort config can now be triggered by a command line argument instead. I have also made the config not COLO-specific (a minimal change). This doesn't work entirely as I expected beforehand, since half of the fields in this cohort config are ignored (see [OPR-44](https://hartwigmedical.atlassian.net/browse/OPR-44)) but the version in this branch works slightly better than the version in the master branch, which should be good enough to merge it, I think.

Let me know what you think,
David


DFKoetsier:
I got permission to merge from Korneel and Lieke through Slack.

PR #358, opened 2023-01-11
Dev 2919: PEACH JAVA review

DFKoetsier:
Hi Paul,

As you asked, I have created a pull request to make the code review more efficient.
This code is not yet ready to be merged.

Thanks for taking a look at it!
David


peach/src/main/java/com/hartwig/hmftools/peach/LiftoverService.java
```diff
@@ -0,0 +1,205 @@
...
+                Stream<String> lines = Files.lines(Paths.get(config.chainFile));
+                PrintWriter pw = new PrintWriter(config.getAdjustedChainFilePath(), StandardCharsets.UTF_8)
+        )
+        {
+            lines.forEachOrdered(line-> pw.println(getAdjustedChainFileLine(line)));
+        }
+        catch (IOException e)
+        {
+            PCH_LOGGER.error("could not create adjusted chain file: ");
+            e.printStackTrace();
```
pauldwolfe, line 73: here you can wrap and rethrow `new RuntimeException(e)` and then exit/print stack in a central place.

peach/src/main/java/com/hartwig/hmftools/peach/PeachUtils.java
```diff
@@ -0,0 +1,15 @@
+package com.hartwig.hmftools.peach;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class PeachUtils
+{
+    public static final Logger PCH_LOGGER = LogManager.getLogger(PeachApplication.class);
```
pauldwolfe, line 8: Just FYI using a central logger this way does have the downside that you cannot tune logging based on class/package, and you will not see what class logs the message in your log entries.  Nothing wrong with what you've done, just wanted to mention.

peach/src/main/java/com/hartwig/hmftools/peach/data_loader/HaplotypeEventLoader.java
```diff
@@ -0,0 +1,85 @@
...
+        Map<String, Integer> eventIdToCount = new HashMap<>();
+        try(
+                AbstractFeatureReader<VariantContext, LineIterator> reader = getFeatureReader(
+                        vcf, new VCFCodec(), false)
+        )
+        {
+            for(VariantContext variantContext : reader.iterator())
+            {
+                if(variantContext.isFiltered())
+                    continue;
```
pauldwolfe, line 35: using continues this way can lead to confusion and occasionally bugs.  I would prefer to just use brackets and either nested or combined clauses in if statements.

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        Chromosome chromosome = HumanChromosome.fromString(splitEventId[1]);
+        int position = Integer.parseInt(splitEventId[2]);
+        String ref = splitEventId[3];
+        String alt = splitEventId[4];
+        VariantHaplotypeEvent event = new VariantHaplotypeEvent(chromosome, position, ref, alt);
+        if (!event.id().equals(eventId))
+        {
+            String error_msg = String.format(
+                    "VariantHaplotypeEvent derived from ID ID '%s' has different ID '%s'",
```
mdenburger, line 65: Did you mean "event ID" instead of "ID ID"?

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
```diff
@@ -0,0 +1,124 @@
...
+    public static VariantHaplotypeEvent fromVariantContext(VariantContext variantContext)
+    {
+        Chromosome chromosome = HumanChromosome.fromString(variantContext.getContig());
+        int position = variantContext.getStart();
+        String ref = variantContext.getReference().getBaseString();
+        List<Allele> alts = variantContext.getAlternateAlleles();
+        if(alts.size() > 1)
+        {
+            String error_msg = String.format(
+                    "Cannot handle variant with multiple alts: %s:%s%s>?",
```
mdenburger, line 82: Is the ">?" suffix intentional?

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
```diff
@@ -0,0 +1,124 @@
...
+    public String toString(){
+        return id();
+    }
+
+    @NotNull
+    public String id()
+    {
+        return new StringJoiner(HaplotypeEvent.EVENT_ID_DELIMITER)
+                .add(EVENT_TYPE_STRING)
+                .add(String.format("chr%1$s", chromosome))
```
mdenburger, line 100: Since there's only one argument, the position part `1$` is redundant

peach/src/main/java/com/hartwig/hmftools/peach/haplotype/Haplotype.java
```diff
@@ -0,0 +1,5 @@
+package com.hartwig.hmftools.peach.haplotype;
+
+public interface Haplotype
```
mdenburger, line 3: This interface feels strange. Both implementations (`DefaultHaplotype` and `NonDefaultHaplotype`) basically have the same properties (`name`, `events` and `isWildType`), the only difference is interpretation of the events. 

I would move at least `name` and `isWildType` to the interface. That way you don't need to cast to access these fields.

peach/src/main/java/com/hartwig/hmftools/peach/haplotype/HaplotypeCombination.java
```diff
@@ -0,0 +1,68 @@
...
+{
+    @NotNull
+    private final Map<String, Integer> haplotypeNameToCount;
+
+    public HaplotypeCombination(@NotNull Map<String, Integer> haplotypeNameToCount)
+    {
+        if (haplotypeNameToCount.values().stream().anyMatch(c -> c <= 0))
+        {
+            String error_msg = String.format(
+                    "Invalid haplotype combination with non-positive count(s): %s", haplotypeNameToCount
```
mdenburger, line 20: How can counts become negative?

peach/src/main/java/com/hartwig/hmftools/peach/event/HaplotypeEvent.java
```diff
@@ -0,0 +1,13 @@
+package com.hartwig.hmftools.peach.event;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface HaplotypeEvent
```
mdenburger, line 5: There's only one haplotype event right now (the `VariantHaplotypeEvent`). I guess there will be more types in the future?

mdenburger (changes requested):
Noted some small improvements here and there, nothing big. I wonder about the possibility of negative counts, that seems odd.

peach/src/main/java/com/hartwig/hmftools/peach/PeachApplication.java
```diff
@@ -0,0 +1,108 @@
...
+        {
+            if (!config.isValid())
+            {
+                throw new IllegalArgumentException("invalid config");
+            }
+            PCH_LOGGER.info("running PEACH");
+
+            PCH_LOGGER.info("creating output directory");
+            // Output directory cannot be null in valid config
+            assert config.outputDir != null;
```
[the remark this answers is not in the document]
DFKoetsier: I did, but IntelliJ's type checker doesn't get that. The assert and comment were meant to make clear to both the type checker and the reader that this was not a possible state, but I would be fine with removing this or changing the comment.

peach/src/main/java/com/hartwig/hmftools/peach/PeachApplication.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        try
+        {
+            if (!config.isValid())
+            {
+                throw new IllegalArgumentException("invalid config");
+            }
+            PCH_LOGGER.info("running PEACH");
+
+            PCH_LOGGER.info("creating output directory");
```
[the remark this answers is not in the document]
DFKoetsier: Agreed

peach/src/main/java/com/hartwig/hmftools/peach/HaplotypeCaller.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private List<List<NonDefaultHaplotype>> getPossibleNonDefaultHaplotypes(
+            Map<String, Integer> eventIdToUnexplainedCount,
+            List<NonDefaultHaplotype> candidateHaplotypes)
+    {
+        // Use recursive descent to efficiently go through all possibilities
+        if (eventIdToUnexplainedCount.values().stream().anyMatch(c -> c < 0))
+        {
+            return new ArrayList<>();
```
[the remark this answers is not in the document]
DFKoetsier: Agreed

peach/src/main/java/com/hartwig/hmftools/peach/HaplotypeCaller.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private Map<String, Integer> getEventIdToUnexplainedCountAfterTry(
+            Map<String, Integer> eventIdToUnexplainedCount, NonDefaultHaplotype haplotypeToTry
+    )
+    {
+        HashMap<String, Integer> newEventIdToCount = Maps.newHashMap(eventIdToUnexplainedCount);
+        for (HaplotypeEvent event : haplotypeToTry.events)
+        {
+            int newCount = newEventIdToCount.getOrDefault(event.id(), 0) - 1;
```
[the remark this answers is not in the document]
DFKoetsier: It was intended. Combinations with negative counts get rejected at the start of the recursive method, so all combinations that result in negative counts were (correctly) dismissed.

I agree that this was confusing, so I have refactored this code to reject impossible haplotypes before calculating the "newEventIdToCount".

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
DFKoetsier: Yes I did, thanks

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
DFKoetsier: It was. Variants are usually denoted as something along the lines 5:1638G>A, which means that on chromosome 5 at position 1638 the G in the reference genome is replaced by an A. I intended this error message to say something like "5:1638G>?" to show the relevant information without having to show all of the relevant alts.

I have added some quotes and replaced the question mark with three dots to make the message look more intentional.

peach/src/main/java/com/hartwig/hmftools/peach/event/VariantHaplotypeEvent.java
DFKoetsier: Agreed

peach/src/main/java/com/hartwig/hmftools/peach/haplotype/Haplotype.java
DFKoetsier: I have moved name and isWildType to the interface

peach/src/main/java/com/hartwig/hmftools/peach/haplotype/HaplotypeCombination.java
DFKoetsier: They shouldn't be able to. I would only be possible if some kind of bug was at play. I just put this in here to make sure such bugs would be caught before creating output files.

Would it be better to skip such tests/checks?

peach/src/main/java/com/hartwig/hmftools/peach/event/HaplotypeEvent.java
DFKoetsier: Yes, there will be also be events for deletions and likely fusions. But this is out of scope for the first Peach Java release.

kzuberihmf (approved):
Hi David, had a look at the changes to Orange and Orange-datamodel and they look fine to me and seem to match the description you provided.

One tiny possible loose end is that there is a test file in orange test_run/peach, I assume it could be renamed to the new extension. Though this could also be done when support for the old peach is removed.

DFKoetsier:
> Hi David, had a look at the changes to Orange and Orange-datamodel and they look fine to me and seem to match the description you provided.
> 
> One tiny possible loose end is that there is a test file in orange test_run/peach, I assume it could be renamed to the new extension. Though this could also be done when support for the old peach is removed.

Great, thanks! I plan to change this once I have put Java Peach into the pipeline.

PR #381, opened 2023-03-28
ACTIN-24: Use one instance of ReportResources to avoid duplicating fonts

jbartletthmf:
Also maximize PDF compression, though this didn't improve test file
size.

Fonts are now final and accessed non-statically, which avoids
concurrency concerns that could arise with a singleton. Test report
size was reduced by 28% when reusing fonts.


orange/src/main/java/com/hartwig/hmftools/orange/report/ReportResources.java
```diff
@@ -46,66 +48,81 @@ public final class ReportResources {
...
+    public Cells cells() {
+        return new Cells(this);
     }
 
-    @NotNull
-    public static PdfFont fontBold() {
-        // Cannot be created statically as every PDF needs their own private font objects.
-        return createFontFromProgram(loadFontProgram(FONT_BOLD_PATH));
+    public Tables tables() {
+        return new Tables(this);
```
kduyvesteyn, line 77: See comment for Cells

orange/src/main/java/com/hartwig/hmftools/orange/report/ReportResources.java
kduyvesteyn, line 1: Class should be non-final, no longer a utility class

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/FrontPageChapter.java
```diff
@@ -126,50 +129,35 @@ private String purpleQCString() {
...
+                Maps.immutableEntry("Tumor mutational load:", tmlString()),
+                Maps.immutableEntry("HR deficiency score:", hrDeficiencyString()),
+                Maps.immutableEntry("DPYD status:", dpydStatus()),
+                Maps.immutableEntry("Number of SVs:", svTmbString()),
+                Maps.immutableEntry("Max complex cluster size:", maxComplexSizeString()),
+                Maps.immutableEntry("Telomeric SGLs:", telomericSGLString()),
+                Maps.immutableEntry("Number of LINE insertions:", lineCountString())).forEach(entry -> {
+            summary.addCell(cells.createKey(entry.getKey()));
+            summary.addCell(cells.createValue(entry.getValue()));
+        });
```
kduyvesteyn, line 160: Impressive :)

orange/src/main/java/com/hartwig/hmftools/orange/report/util/Cells.java
```diff
@@ -15,80 +15,83 @@
 
 public final class Cells {
```
kduyvesteyn, line 16: Should no longer be final

kduyvesteyn (commented):
Some general thoughts:
 - Have no super strong opinion but I'd think reportResources should be a property of some classes rather than an input to the main function. Eg, the classes in "tables" package are not fundamentally different from Cells and Tables. In theory a pdf could generate the same table multiple times (with different inputs). The reportResources would be a property of the class generating the tables (exactly similar to Cells and Tables)
 - I would still suggest to retain the IntelliJ annotations. They are quite widely used in hmftools so ORANGE would become inconsistent if we start removing them.

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/GermlineFindingsChapter.java
```diff
@@ -179,16 +180,16 @@ private void addGermlineCNAberrations(@NotNull Document document) {
...
         }
     }
 
-    private void addPharmacogenetics(@NotNull Document document) {
+    private void addPharmacogenetics(@NotNull Document document, @NotNull ReportResources reportResources) {
         Set<PeachGenotype> peach = report.peach();
         if (peach != null) {
             String titlePharmacogenetics = "Pharmacogenetics (" + peach.size() + ")";
-            document.add(PharmacogeneticsTable.build(titlePharmacogenetics, contentWidth(), peach));
+            document.add(PharmacogeneticsTable.build(titlePharmacogenetics, contentWidth(), peach, reportResources));
```
jbartletthmf, line 192: @kduyvesteyn, I'll reply here for context and to facilitate discussion since github won't let me reply to your general comment.
> * Have no super strong opinion but I'd think reportResources should be a property of some classes rather than an input to the main function. Eg, the classes in "tables" package are not fundamentally different from Cells and Tables. In theory a pdf could generate the same table multiple times (with different inputs). The reportResources would be a property of the class generating the tables (exactly similar to Cells and Tables)

I started out by providing the resources to the `tables` classes in the constructor, but each of these classes has a private, no-argument constructor and all interactions are with the static `build` function. Taking your suggestion would mean changing all calls to build tables as follows:
```suggestion
            document.add(new PharmacogeneticsTable(reportResources).build(titlePharmacogenetics, contentWidth(), peach));
```
I find that clunkier to work with, and I don't think making the `ReportResources` a class property would improve much in the rare case where we call the same table class multiple times since we're just passing a reference either way.

On the other hand, adding a `ReportResources` property to the `ReportChapter` classes would eliminate the need to pass that parameter around through so many method calls, so I'm happy to refactor on that side.

orange/src/main/java/com/hartwig/hmftools/orange/report/ReportResources.java
```diff
@@ -46,66 +48,81 @@ public final class ReportResources {
...
         // To make sure every decimal format uses a dot as separator rather than a comma.
         return new DecimalFormat(format, DecimalFormatSymbols.getInstance(Locale.ENGLISH));
     }
 
-    @NotNull
-    public static PdfFont fontRegular() {
-        // Cannot be created statically as every PDF needs their own private font objects.
-        return createFontFromProgram(loadFontProgram(FONT_REGULAR_PATH));
+    public Cells cells() {
+        return new Cells(this);
```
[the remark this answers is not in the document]
jbartletthmf: I was thinking that `Cells` is a kind of resource that we could access through our main resource object, but I'm fine with some kind of `Cells.createFrom(resources)` approach as well.

orange/src/main/java/com/hartwig/hmftools/orange/report/ReportResources.java
```diff
@@ -46,66 +48,81 @@ public final class ReportResources {
...
 
-    @NotNull
-    public static PdfFont fontBold() {
-        // Cannot be created statically as every PDF needs their own private font objects.
-        return createFontFromProgram(loadFontProgram(FONT_BOLD_PATH));
+    public Tables tables() {
+        return new Tables(this);
+    }
+
+    public PdfFont fontBold() {
```
jbartletthmf, line 80: @kduyvesteyn, again replying here for ease of conversation:

> * I would still suggest to retain the IntelliJ annotations. They are quite widely used in hmftools so ORANGE would become inconsistent if we start removing them.

I was trying to retain the annotations as I made my changes, but it looks like I missed adding some `@NotNull` annotations when I added to this file. Is that what you were referring to?

orange/src/main/java/com/hartwig/hmftools/orange/report/ReportResources.java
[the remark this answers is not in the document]
jbartletthmf: The constructor for `Cells` now takes a `ReportResources` parameter, so I'll use that.

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/GermlineFindingsChapter.java
kduyvesteyn: Changing the table classes would indeed not have that much impact. I also considered proposing changing the Chapter interface but then I think a Document is always tied to one ReportResources, so I think at least the interface now is consistent. If we make `reportResources` a property then maybe Document should be one as well, since we are not supposed to generate 2 different documents with one `reportResources`. 

In any case I'd also be fine with current approach. If you feel that's ok I suggest to leave as you initially made it.

kduyvesteyn (approved):
Looks all good to me.

kduyvesteyn (approved):

PR #382, opened 2023-03-28
ACTIN-32: Improve formatting for undetermined HRD type

jbartletthmf:


kduyvesteyn (approved):
Looks all good. 

Note that we could consider changing 1hrdType` to an enum. The developer of CHORD (Luan Nguyen) has recently joined HMF so feel free to ask to him, the list of values is probably quite stable. Also fine to keep as string for now, no actual logic depends on it (apart from formatting as string in report)

kduyvesteyn (approved):

pauldwolfe (approved):

PR #385, opened 2023-04-04
ACTIN-41: Add optional disclaimer to ORANGE

jbartletthmf:


orange/src/main/java/com/hartwig/hmftools/orange/report/PageEventHandler.java
```diff
@@ -67,7 +68,7 @@ void resetChapterPageCounter() {
     }
 
     void writeTotalPageCount(@NotNull PdfDocument document) {
```
kduyvesteyn, line 70: Not related to this actual change, but shouldn't this function be called `writeFooters`? Your renames in this class are good renames but suspect you missed one :)

orange/src/main/java/com/hartwig/hmftools/orange/report/ReportWriterFactory.java
```diff
@@ -12,11 +12,11 @@ private ReportWriterFactory() {
...
     public static ReportWriter createToDiskWriter(@NotNull OrangeConfig config) {
         String outputDir = config.outputDir();
-        return new ReportWriter(true, outputDir, new PlotPathResolver(outputDir));
+        return new ReportWriter(true, outputDir, new PlotPathResolver(outputDir), config.addDisclaimer());
     }
 
     @NotNull
     public static ReportWriter createInMemoryWriter() {
-        return new ReportWriter(false, null, new PlotPathResolver(null));
+        return new ReportWriter(false, null, new PlotPathResolver(null), false);
```
kduyvesteyn, line 20: Minor detail but would set this the disclaimer to `true`. the in-memory writer should have everything enabled if possible, leads to a better test (e.g. to capture a null-pointer in the section enabled by the disclaimer)

orange/src/main/java/com/hartwig/hmftools/orange/report/components/Footer.java
```diff
@@ -10,56 +10,66 @@
...
+        PdfFormXObject pageNumberTemplate = new PdfFormXObject(new Rectangle(0, 0, 450, 20));
         canvas.addXObject(pageNumberTemplate, 58, 20);
-        pageNumberTemplates.add(new PageNumberTemplate(pageNumber, pageNumberTemplate));
+        footerTemplates.add(new FooterTemplate(pageNumber, pageNumberTemplate, addDisclaimer));
 
         canvas.release();
     }
 
-    public void writeTotalPageCount(@NotNull PdfDocument document) {
+    public void writePageCounts(@NotNull PdfDocument document) {
```
kduyvesteyn, line 40: Similar comment as for `PageEventHandler`: Shouldn't this function be called `writeFooters`?

kduyvesteyn (commented):
General comment: I noticed you have not always included IntelliJ annotations, and think I will stop commenting on that. Maybe I should simply accept the fate of annotations also in hmftools :)

Also, don't forget to update the README "upcoming changes" section

kduyvesteyn (approved):
All changes look good.

One example of where I would have used NotNull is with the Footer reportResources property, but I am also fine to be less strict on this going forward. You shouldn't feel burdened by maintaining a deprecated way of avoiding null pointer exceptions :)

jbartletthmf:
I missed that `ReportResources` because I added it in a previous PR. I'll add the annotations now because we don't currently have a better way of avoiding NPEs in Java.

PR #402, opened 2023-05-12
DEV-3251 Fix the orange conversion for virus interpretation

pauldwolfe:
Add the two missed fields and also handle the nullable interpretation field.

Adds a test as well which asserts for an empty, minimal, and exhaustive test datamodel. If we like how this test looks will add the same for the other converted datamodel classes.

### DIFF
--- a/orange/src/test/java/com/hartwig/hmftools/orange/conversion/OrangeConversionTest.java
+++ b/orange/src/test/java/com/hartwig/hmftools/orange/conversion/OrangeConversionTest.java
@@ -1,0 +1,46 @@
+package com.hartwig.hmftools.orange.conversion;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
+
+import com.hartwig.hmftools.datamodel.virus.AnnotatedVirus;
+import com.hartwig.hmftools.datamodel.virus.VirusInterpreterData;
+
+import org.junit.Test;
+
+public class OrangeConversionTest {
+
+    @Test
+    public void shouldConvertAnnotatedVirus() {
+        assertTrue(OrangeConversion.convert(TestCommonDatamodelFactory.emptyVirusInterpreterData()).allViruses().isEmpty());
+        assertEqualsValue(OrangeConversion.convert(TestCommonDatamodelFactory.minimalVirusInterpreterData()),
+                TestCommonDatamodelFactory.minimalVirusInterpreterData());
+        assertEqualsValue(OrangeConversion.convert(TestCommonDatamodelFactory.exhaustiveVirusInterpreterData()),
+                TestCommonDatamodelFactory.exhaustiveVirusInterpreterData());
+    }
+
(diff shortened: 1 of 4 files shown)


orange/src/test/java/com/hartwig/hmftools/orange/conversion/OrangeConversionTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    public void shouldConvertAnnotatedVirus() {
+        assertTrue(OrangeConversion.convert(TestCommonDatamodelFactory.emptyVirusInterpreterData()).allViruses().isEmpty());
+        assertEqualsValue(OrangeConversion.convert(TestCommonDatamodelFactory.minimalVirusInterpreterData()),
+                TestCommonDatamodelFactory.minimalVirusInterpreterData());
+        assertEqualsValue(OrangeConversion.convert(TestCommonDatamodelFactory.exhaustiveVirusInterpreterData()),
+                TestCommonDatamodelFactory.exhaustiveVirusInterpreterData());
+    }
+
+    private static void assertEqualsValue(VirusInterpreterData converted, com.hartwig.hmftools.common.virus.VirusInterpreterData input) {
```
jbartletthmf, line 24: I think it's more conventional to provide the expected value as the first parameter. The effect is mostly the same, but it can impact failure messages.

orange/src/test/java/com/hartwig/hmftools/orange/conversion/OrangeConversionTest.java
```diff
@@ -0,0 +1,46 @@
...
+        for (int i = 0; i < input.size(); i++) {
+            assertEqualsValue(converted.get(i), input.get(i));
+        }
+    }
+
+    private static void assertEqualsValue(AnnotatedVirus converted, com.hartwig.hmftools.common.virus.AnnotatedVirus input) {
+        assertEquals(converted.name(), input.name());
+        assertEquals(converted.qcStatus().name(), input.qcStatus().name());
+        assertEquals(converted.integrations(), input.integrations());
+        assertEquals(converted.percentageCovered(), input.percentageCovered(), 0.01);
```
jbartletthmf, line 40: The allowed error is typically set in an `EPSILON` constant in ORANGE tests.

orange/src/test/java/com/hartwig/hmftools/orange/conversion/TestCommonDatamodelFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+class TestCommonDatamodelFactory {
+
+    static VirusInterpreterData emptyVirusInterpreterData() {
+        return ImmutableVirusInterpreterData.builder().build();
+    }
+
+    static VirusInterpreterData minimalVirusInterpreterData() {
+        AnnotatedVirus virus = minimalAnnotatedVirusBuilder().build();
+        return ImmutableVirusInterpreterData.builder().addReportableViruses(virus).addAllAllViruses(List.of(virus)).build();
```
jbartletthmf, line 20: Do you need `addAllAllViruses()` here or could you just provide `virus` to `addAllViruses()` without having to build a list?

jbartletthmf (approved):

PR #412, opened 2023-06-08
ACTIN-151: Fix expression selection bug for unknown cohort

jbartletthmf:
The cohort isn't indicated when ingesting the isofox results but we can detect the special n/a value of -1. When this isn't provided, there is no filtering on cohort percentile so the number of genes meeting the expression thresholds is expected to be higher.

### DIFF
--- a/hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
+++ b/hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
@@ -20,6 +20,7 @@ import org.apache.logging.log4j.Logger;
 public class GeneExpressionDistributionData
 {
     private static final Logger LOGGER = LogManager.getLogger(GeneExpressionDistributionData.class);
+    public static final int NOT_AVAILABLE = -1;
 
     private final Map<String, Map<String, double[]>> mGenePercentiles; // by geneId and then cancer type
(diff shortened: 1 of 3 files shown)


hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
```diff
@@ -20,6 +20,7 @@
 public class GeneExpressionDistributionData
 {
     private static final Logger LOGGER = LogManager.getLogger(GeneExpressionDistributionData.class);
+    public static final int NOT_AVAILABLE = -1;
```
pauldwolfe, line 23: can be package private i think

pauldwolfe (approved):

hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
kduyvesteyn: I am pretty sure it cannot!!

kduyvesteyn (approved):

hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
pauldwolfe: indeed, disregard

PR #417, opened 2023-06-19
ACTIN-170 Persist/read the config with the orange record

pauldwolfe:
Persist the full input config with the orange record such that it can be read back and used when rerunning.

### DIFF
--- a/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeConfig.java
+++ b/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeConfig.java
@@ -1,0 +1,132 @@
+package com.hartwig.hmftools.datamodel.orange;
+
+import java.time.LocalDate;
+import java.util.Set;
+
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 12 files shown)


jbartletthmf (commented):
The approach makes sense to me.

kduyvesteyn (commented):
The approach doesn't make sense to me. Most of the fields in OrangeConfig will be useless since they refer to file paths as they existed on the VM running ORANGE. 

I would propose to take a different route:
 - In bash when we need the experimentDate, we either extract from LIMS or from an existing ORANGE json file.
 - In bash when we need the tumor DOIDs, we either extract from sampleAPI or from an existing ORANGE json file.

this way ORANGE doesn't actually depend on ORANGE and we don't expose all kinds of runtime information in a datamodel that is now purely data.

pauldwolfe:
Yeah fair enough, I guess my idea was that as a principle its nice to simply capture inputs and store them immutably with the outputs for a rerun. The file paths do spoil the party. What I wanted to avoid was some brittle `jq` in a script somewhere plucking dates and doids out of JSON files... but perhaps that is the easiest and most practical way.

PR #424, opened 2023-07-06
Dev 3451

charlesshale:
I've tested Orange in the pipeline using these new config options and the report output is identical.


orange/src/main/java/com/hartwig/hmftools/orange/OrangeApplication.java
```diff
@@ -25,18 +26,21 @@ public class OrangeApplication {
...
+        }
 
-        OrangeConfig config = null;
         try {
-            config = OrangeConfig.createConfig(new DefaultParser().parse(options, args));
+            OrangeConfig config = OrangeConfig.createConfig(configBuilder);
+            new OrangeApplication(config).run();
         } catch (ParseException exception) {
-            LOGGER.warn(exception);
-            new HelpFormatter().printHelp(APPLICATION, options);
```
pauldwolfe, line 35: Does the help text still get printed somewhere?

orange/src/main/java/com/hartwig/hmftools/orange/OrangeRNAConfig.java
```diff
@@ -19,93 +28,76 @@ public interface OrangeRNAConfig {
...
-        if (!hasCompleteRnaConfig(cmd,
-                RNA_SAMPLE_ID,
-                ISOFOX_GENE_DISTRIBUTION_CSV,
-                ISOFOX_ALT_SJ_COHORT_CSV,
-                ISOFOX_SUMMARY_CSV,
-                ISOFOX_GENE_DATA_CSV,
-                ISOFOX_FUSION_CSV,
-                ISOFOX_ALT_SPLICE_JUNCTION_CSV)) {
-            LOGGER.debug("No proper RNA input fed to ORANGE");
+    static OrangeRNAConfig createConfig(final ConfigBuilder configBuilder) {
```
pauldwolfe, line 53: Guess this should be annotated @Nullable? Or is that not part of the current hmftools code standard?

orange/src/main/java/com/hartwig/hmftools/orange/OrangeRNAConfig.java
```diff
@@ -19,93 +28,76 @@ public interface OrangeRNAConfig {
...
+
+        if(!anyConfigPresent) {
+            LOGGER.info("RNA config not present");
+            return null;
+        }
+
+        boolean allConfigPresent = hasRnaSampleId && hasIsofoxDir && hasGeneDistribution && hasAltSjCohortFreq;
+
+        if(!allConfigPresent) {
+            LOGGER.warn("RNA missing required config items: rnaSampleId({}) isofoxDir({}) geneCohort({}) altSjCohort({})",
```
pauldwolfe, line 70: Maybe this should be an ERROR/faillure? Feels like if you've specified any of these arguments you probably intend to run RNA.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeApplication.java
charlesshale: Yes it does by the ConfigBuilder class

orange/src/main/java/com/hartwig/hmftools/orange/OrangeRNAConfig.java
charlesshale: It shouldn't be @Nullable, no - it is correct as-is.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeRNAConfig.java
charlesshale: I kept with existing behaviour in Orange here.

PR #425, opened 2023-07-06
DEV-3374 include buildscript and cloudbuild 

KasperWolsink99:


cloudbuild.yaml
```diff
@@ -0,0 +1,9 @@
+steps:
+  - id: 'Tag, release, build'
+    name: 'eu.gcr.io/hmf-build/jdk-mvn-python:1.2.0-alpha.2'
```
KasperWolsink99, line 3: TODO, change this to stable release

pauldwolfe (approved):

hmftools-build.py
```diff
@@ -0,0 +1,67 @@
+import re
+import subprocess
+import sys
+import xml.etree.ElementTree as ET
+
+# Check if a semver version is included as argument.
+if len(sys.argv) != 2:
+    print(f'Invalid arguments. Usage: {sys.argv[0]} semver-version')
```
mvanniekerkHartwig, line 8: Put semver-version in `<semver-version>`, convention for printing help message on the command line.

hmftools-build.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Set version of module in parent pom.
+subprocess.run(
+    ['mvn', '-f', 'pom.xml', 'versions:set-property', '-DgenerateBackupPoms=false', f'-Dproperty={module}.version',
+     f'-DnewVersion={tag}'],
+    check=True)
+
+# Set version of hmftools dependencies in parent pom.xml
+for hmf_dep in hmftools_dependencies:
+    subprocess.run(['mvn', '-f', 'pom.xml', 'versions:set-property', '-DgenerateBackupPoms=false',
```
mvanniekerkHartwig, line 45: I don't know if you can meaningfully clean up these subprocess run calls, but you can maybe extract the call to a function?

mvanniekerkHartwig (changes requested):

KasperWolsink99:
A thing to consider is that we want access to the actual Jar files. Currently all artifacts are deployed to artifact registry which does not provide an easy way to download the jars. One way to still do it is by running this script:

`curl --oauth2-bearer <gcp-access-token> -o <jar-you-want-to-download>.jar -L`

pauldwolfe (approved):

mvanniekerkHartwig (approved):

PR #433, opened 2023-07-11
DEV-3451 Ensure that mandatory files passed to Orange exist

pauldwolfe:
Update config builder setup to use PATH appropriately and required.

Added checks to all paths derived from directories, previously declared as "nonOptional". The checks will
throw if the file does not exist.

Also throw when input is incomplete for RNA rather than returning null and skipping.

Add a warning when plot dir exists and is non-empty.


kduyvesteyn (changes requested):
Have made a few refactorings but in general it seems like there are still quite some checks missing. E.g. ORANGE doesn't throw an exception if `linxSomaticDir` doesn't exist I believe, and all the resources (e.g. `cohortMappingTsv`) aren't checked either yet.

pauldwolfe:
> Have made a few refactorings but in general it seems like there are still quite some checks missing. E.g. ORANGE doesn't throw an exception if `linxSomaticDir` doesn't exist I believe, and all the resources (e.g. `cohortMappingTsv`) aren't checked either yet.

@kduyvesteyn Any config item configured with `ConfigBuilder.addPath` get checked by the ConfigBuilder framework for existence. It's only the implied files we need to check ourselves. In other words, non of the directories or resources need to be checked, as long as they configured as paths.

PR #434, opened 2023-07-13
DEV-3317 add non-canonical transcripts data to orange

KasperWolsink99:
- Non-cannonical transcripts are now loaded in the `PurpleVariant`s `otherImpacts` field.
- Move `PurpleData` and `PurpleDataLoader` from hmf-common to orange.
- Data is now loaded from a VCF file by `PurpleVariantContextLoader` to a `PurpleVariantContext` object.
- `PurpleVariantContext` is then converted to `PurpleVariant` by `PurpleVariantFactory::fromPurpleVariantContext`


hmf-common/src/main/java/com/hartwig/hmftools/common/variant/impact/VariantTranscriptImpact.java
```diff
@@ -79,6 +82,12 @@ public static List<VariantTranscriptImpact> fromVariantContext(final VariantCont
     public static VariantTranscriptImpact fromVcfData(final String data)
     {
         String[] items = data.split("\\" + VAR_TRANS_IMPACT_ITEM_DELIM, -1);
+
+        // dirty fix: append items if it is less than 7 to prevent out of bound exceptions
```
KasperWolsink99, line 86: Some tests (for example those in orange) have entires where this array has fewer entries than 7. In those cases, the tests would fail with an index out of bounds exception.

I lack the domain knowledge to determine if this is a sign of a more serious problem with the data, or if this is expected sometimes. For now I implemented this fix, but maybe it is worth looking into how to handle these situations.

orange-datamodel/pom.xml
```diff
@@ -42,5 +42,9 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
mvanniekerkHartwig, line 45: orange-datamodel should not have a dependency on hmf-common.

KasperWolsink99: The dependency was added because PurpleVariant contained a list of `VariantTranscriptImpact`'s (which is a class of `hmf-common`).

However, I made some changes that added the `PurpleVariantTranscriptImpact` class and the `VariantTranscriptImpact` instances are mapped to this class. Hence the dependency was can be removed entirely.

hmf-common/src/main/java/com/hartwig/hmftools/common/variant/impact/VariantTranscriptImpact.java
```diff
@@ -12,8 +13,7 @@
 import htsjdk.variant.vcf.VCFHeaderLineType;
 import htsjdk.variant.vcf.VCFInfoHeaderLine;
 
-public class VariantTranscriptImpact
-{
```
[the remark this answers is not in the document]
KasperWolsink99: Fixed the styling issues

kduyvesteyn (changes requested):
The approach to directly load the data from the VCF inside ORANGE looks good to me, but have some quite general comments:
 - I would maintain strict separation between loading and interpreting the data. This is because the interpretation will be pushed upstream in the pipeline at some point, and by keeping it isolated it is more clear what ORANGE adds to the output of e.g. purple which should be produced by purple. So the paveAlgo should not be part of the PurpleVariantFactory but rather keep an intermediate datamodel that loads the data as it exists in the VCF, and passes it into PurpleInterpreter.
 - The PurpleVariantTranscriptImpact object feels duplicate with PurpleTranscriptImpact. Why don't you just populate the otherImpacts with the impacts on transcripts that are non-canonical? ACTIN uses the otherImpacts and requires every field to be populated, so we wouldn't be able to use you new class even. 
 - It seems like Matthijs has made comments on formatting but I believe some things are not consistent with general hmftools code style (I am not 100% sure on those, but when in doubt please check with Charles).

kduyvesteyn:
1) You now create PurpleVariant but don't populate it exactly. I would still create a separate datamodel.

2) The `otherImpacts` are to capture all impacts other than the canonical, so your new field is unnecessary.

3) Every transcript impact has N effects which are interpreted as 1 coding effect (see VariantImpactBuilder.determineCodingEffect).

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -28,135 +23,102 @@
...
     @Test
-    public void canHandleNullVariants() {
-        PurpleVariantFactory factory = createTestFactory();
-        assertNull(factory.create(null));
-    }
+    public void testCanCreateVariantFromContext()
+    {
+        var otherImpacts = List.of(
+                new VariantTranscriptImpact(
+                        "[ENSG00000109265",
```
kduyvesteyn, line 36: I think your logic should filter out these brackets at an early stage. I don't think your own `PurpleVariantContext` should ever contain brackets so you also don't have to test the conversion with brackets. If your `PurpleVariantContext` can contains brackets, you should deal with those in the loader.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantContextLoader.java
```diff
@@ -0,0 +1,167 @@
...
+                .gene(variantImpact.CanonicalGeneName)
+                .ref(contextDecorator.ref())
+                .alt(contextDecorator.alt())
+                .canonicalTranscript(variantImpact.CanonicalTranscript)
+                .canonicalEffect(variantImpact.CanonicalEffect)
+                .canonicalCodingEffect(variantImpact.CanonicalCodingEffect)
+                .canonicalHgvsCodingImpact(variantImpact.CanonicalHgvsCoding)
+                .canonicalHgvsProteinImpact(variantImpact.CanonicalHgvsProtein)
+                .worstCodingEffect(variantImpact.WorstCodingEffect)
+                .otherImpacts(variantTranscriptImpacts)
```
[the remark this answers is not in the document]
KasperWolsink99: I did go for the first option where the canonical impact is filtered out as it seemed like the change that required the least amount of changes

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
KasperWolsink99: Good suggestion and I agree. The inclusion of the brackets does seem unintentional and in my opinion it ideally should be fixed in the `VariantTranscriptImpact` class itself. Maybe I can discuss this with Charles some time.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantContextLoader.java
```diff
@@ -0,0 +1,167 @@
...
+                .rnaDepth(rnaDepth)
+                .adjustedCopyNumber(contextDecorator.adjustedCopyNumber())
+                .adjustedVAF(contextDecorator.adjustedVaf())
+                .minorAlleleCopyNumber(contextDecorator.minorAlleleCopyNumber())
+                .variantCopyNumber(contextDecorator.variantCopyNumber())
+                .biallelic(contextDecorator.biallelic())
+                .genotypeStatus(contextDecorator.genotypeStatus(reference))
+                .repeatCount(contextDecorator.repeatCount())
+                .subclonalLikelihood(variantContext.getAttributeAsDouble(SUBCLONAL_LIKELIHOOD_FLAG, 0))
+                .localPhaseSets(variantContext.getAttributeAsIntList(LOCAL_PHASE_SET, 0))
```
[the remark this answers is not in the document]
KasperWolsink99: Made it null of `LOCAL_PHASE_SET` attribute is absent so it alligns more with existing code

orange/src/test/resources/purple/variants.vcf
```diff
@@ -0,0 +1,32 @@
...
+##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
+##FORMAT=<ID=RABQ,Number=2,Type=Integer,Description="Raw allelic base quality">
+##FORMAT=<ID=RAD,Number=2,Type=Integer,Description="Raw allelic depth">
+##FORMAT=<ID=RC_CNT,Number=7,Type=Integer,Description="Read context counts [Full, Partial, Core, Realigned, Alt, Reference, Total]">
+##FORMAT=<ID=RC_IPC,Number=1,Type=Integer,Description="Read context improper pair count">
+##FORMAT=<ID=RC_JIT,Number=3,Type=Integer,Description="Read context jitter [Shortened, Lengthened, QualityPenalty]">
+##FORMAT=<ID=RC_QUAL,Number=7,Type=Integer,Description="Read context quality [Full, Partial, Core, Realigned, Alt, Reference, Total]">
+##FORMAT=<ID=RDP,Number=1,Type=Integer,Description="Raw read depth">
+##FORMAT=<ID=SB,Number=1,Type=Float,Description="Strand bias - percentage of first-in-pair reads">
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	COLO829v003R	COLO829v003T
```
kduyvesteyn, line 31: Minor detail but I generally use proper test names rather than "COLO829v003R". In `test_run` I use `ref_sample` and `tumor_sample` for example.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantContextLoader.java
```diff
@@ -0,0 +1,219 @@
...
+                cleanedGeneName,
+                cleanedTranscript,
+                cleanedEffects,
+                impact.SpliceRegion,
+                cleanedHgvsCoding,
+                cleanedHgvsProtein);
+    }
+
+    @Nullable
+    private static String stripSquareBracketsAndWhiteSpace(String s)
```
kduyvesteyn, line 189: Minor detail but would use `string` rather than `s`

Also, not entirely clear to me whether this logic is explicitly tested somewhere? I would at least test that an input VariantTranscriptImpact with brackets is handled correctly (though might have skipped over it, maybe `testCanReadVcfFile` tests it implicitly?).

kduyvesteyn (approved):
All changes look good to me! 

A few minor comments, feel free to followup or ignore.

PR #445, opened 2023-08-15
ACTIN-261: Add new cohort constants & cause error instead of warn

ninajacobs:


orange/src/main/java/com/hartwig/hmftools/orange/cohort/mapping/CohortConstants.java
```diff
@@ -21,8 +25,29 @@ public final class CohortConstants {
         // Kidney cancer is a tough one to spread out without affecting either Kidney or Urothelial tract itself.
         DOID_COMBINATIONS_TO_MAP_TO_OTHER.add(Sets.newHashSet("263"));
 
-        // Combination of Urethra cancer and renal cell cancer cannot easily be mapped.
+        // Combination of urethra cancer and renal cell cancer cannot easily be mapped.
         DOID_COMBINATIONS_TO_MAP_TO_OTHER.add(Sets.newHashSet("734", "2671", "4450"));
+
+        // Combination of esophagus or stomach cancer & gastroesophageal system cancer should resolve to esophagus or stomach.
+        DOID_COMBINATIONS_TO_MAP_TO_ESOPHAGUS.add(Sets.newHashSet("0080374", "5041"));
+        DOID_COMBINATIONS_TO_MAP_TO_ESOPHAGUS.add(Sets.newHashSet("0080374", "1107"));
```
jbartletthmf, line 33: This solution works, but if we expect the number of cases to keep growing we may want to generalize it with a map like:
```java
        Map<Set<String>, String> DOID_COMBINATION_MAP = new HashMap<>();
        
        for (Set<String> combination : DOID_COMBINATIONS_TO_MAP_TO_OTHER) {
            DOID_COMBINATION_MAP.add(combination, COHORT_OTHER);
        }
        for (Set<String> combination : DOID_COMBINATIONS_TO_MAP_TO_ESOPHAGUS) {
            DOID_COMBINATION_MAP.add(combination, COHORT_ESOPHAGUS);
        }
        for (Set<String> combination : DOID_COMBINATIONS_TO_MAP_TO_STOMACH) {
            DOID_COMBINATION_MAP.add(combination, COHORT_STOMACH);
        }
```
Then, we could handle all mapping with a single check:
```java
        String mappedCohort = CohortConstants.DOID_COMBINATION_MAP.get(sample.doids());
        if (mappedCohort != null) {
            LOGGER.debug("Mapping {} to {} because of specific doid combination: {}",
                    sample.sampleId(),
                    mappedCohort,
                    sample.doids());
            return mappedCohort;
        }
```

jbartletthmf (approved):

orange/src/main/java/com/hartwig/hmftools/orange/cohort/mapping/CohortConstants.java
ninajacobs: Thanks John! I had to use 'put' instead of 'add' for the map, but I think it works as expected now?

jbartletthmf: Whoops! 🤦‍♂️ I'm glad you understood. Looks good 🙂

PR #446, opened 2023-08-15
ACTIN-266: Bump test data and example report to v5.33 pipeline

kduyvesteyn:
In addition to updating the test data to v5.33 (in /src/test/resources/test_run) and updating the test report to v5.33 (in /src/main/resources/Test.orange.pdf), have done the following:
 - Added a "test_run_ensembl" cache to hold dummy data for all genes that have germline deletions in the test data. ORANGE expects these entries to be there and decided to create a mock cache rather than a mock object. 
 - Fixed a bug when loading up the cache, now also transcripts that are non-ensembl are loaded up. The impact of solving this bug is:
   - If the canonical transcript of a gene is a non-ensembl transcript then ORANGE will crash in case of a germline deletion in that gene
   - Variants on non-ensembl transcripts will always have `affectedCodon` and `affectedExon` set to null
 - Cleaned up a few inconsistencies.


orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -353,6 +353,7 @@ private static EnsemblDataCache loadEnsemblDataCache(@NotNull OrangeConfig confi
     {
         EnsemblDataCache ensemblDataCache = new EnsemblDataCache(config.ensemblDataDirectory(),
                 RefGenomeVersion.from(config.refGenomeVersion().name()));
+        ensemblDataCache.setRequireNonEnsemblTranscripts();
```
jbartletthmf, line 356: Would `includeNonEnsemblTranscripts()` be a better name for this method? Also, is the fix captured by any tests?

jbartletthmf (approved):

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
```diff
@@ -96,19 +102,24 @@ private static ImmutableCuppaPrediction prediction(@NotNull String cancerType, d
...
         Double expectedValue = function.apply(expected);
         Double actualValue = function.apply(actual);
-        if (expectedValue == null) {
+        if(expectedValue == null)
+        {
             assertNull(actualValue);
-        } else {
+        }
+        else
+        {
```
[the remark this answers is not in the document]
kduyvesteyn: Paul has recently decided with Charles to reformat ORANGE code to the hmftools code standards (the xml is included in the hmftools repo), so I've been applying this style to files that I change.

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
kduyvesteyn: The current function name is consistent with similar function names in this object and the object is part of hmf-common so I'd rather not touch that.

`OrangeAlgoTest` fails without this line, because it uses a dummy ensembl cache where none of the transcripts match with the ensembl pattern. Its not explicit but this failing test was the actual reason I found this bug in the first place so at least its tested!

PR #466, opened 2023-09-28
ACTIN-324: add UGT1A1 status to front page of orange report

kzuberihmf:
Small change, adds UGT1A1 directly under DPYD on front page of report in same format.

Testing manually with ReportGeneratorTestApplication, outputs "*1_HOM (Normal Function)" also for this gene.


kduyvesteyn (approved):
Changes look good. 

Would only recommend to add a note to the "upcoming" release notes in the ORANGE README, see near the bottom of this file (and my changes from ACTIN-323 as example)

PR #470, opened 2023-10-04
ACTIN-100: Improve consistency and quality of orange-datamodel

kzuberihmf:
Various improvements as outlined in Jira. Most of these are fairly mechanical refactoring. This will be a breaking change and require a new major version for orange-datamodel.

A couple of notes:

* Not sure if the change here for PurpleFIt.containsTumorCells() considering FAIL_NO_TUMOR qc status matches the intent. Did we want to raise some error if inconsistent with NO_TUMOR fitted purity method? Also I didn't see a test for the interpretation of containsTumorCells(), but we may want to add one (after confirmation that the change here is doing what's intended)

* Didn't remove containsTumorCells or hasSufficientQuality as the ticket description seemed undecided on this.

* Did not integrate RnaQCFilter enum into orange-datamodel, need some more info on this. E.g. if intent was to replace something like PurpleQCStatus, the enums are quite different.


orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeRecord.java
```diff
@@ -31,9 +31,9 @@ public interface OrangeRecord {
     String sampleId();
 
     @NotNull
-    LocalDate experimentDate();
+    LocalDate samplingDate();
 
-    @Nullable
+    @NotNull
     ExperimentType experimentType();
```
pauldwolfe, line 37: @kduyvesteyn is experiment still a good term here? Perhaps SequencingType is more descriptive?

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/HotspotType.java
```diff
@@ -1,6 +1,7 @@
 package com.hartwig.hmftools.datamodel.purple;
 
-public enum Hotspot {
+public enum HotspotType
```
pauldwolfe, line 3: Suggestion, what about `HotspotProximity` with values INSIDE, NEAR, OUTSIDE

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeRecord.java
kduyvesteyn: I discussed with Ewart. The term "experiment" is used in our lab extensively but is an internal term. E.g. when the lab isolates DNA from a biopsy they call this a "DNA isolation experiment". 

I have asked whether SequencingType would be a better name but both targeted and WGS are sequenced identically (just with different libraries), so then SequencingType isn't a better name either. 

I can't think of a better name than ExperimentType at the moment, it relates quite well (but not perfectly) with the naming conventions in the lab

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/HotspotType.java
kduyvesteyn: I like the proposal for the enum name, but the values could be better. A hotspot is never "wider" than a variant so INSIDE doesn't make sense. Maybe MATCH, NEAR, NON-MATCH would be better.. although I don't like the term NON_MATCH. 

In a way this field captures 2 booleans:
 - Does this variant match with a hotspot?
 - If not, is there a hotspot nearby?

Maybe we should model this more explicitly... the benefit of the current enum though is that the 2nd boolean is irrelevant if the first boolean is true.

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeRecord.java
pauldwolfe: Ok I'm satisfied!

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/isofox/RnaFusion.java
```diff
@@ -8,11 +8,17 @@
 @Gson.TypeAdapters
 @Value.Immutable
 @Value.Style(passAnnotations = { NotNull.class, Nullable.class })
-public interface RnaFusion {
-
+public interface RnaFusion
+{
     @NotNull
     String name();
```
kduyvesteyn, line 15: I would make this a derived property `display`, similar to LinxFusion

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/isofox/RnaFusion.java
```diff
@@ -8,11 +8,17 @@
...
 
+    @NotNull
+    String geneStart();
+
+    @NotNull
+    String geneEnd();
+
```
kduyvesteyn, line 21: It's not ideal but I would rephrase this to geneUp / geneDown, since all other fields in RnaFusion are called up/down already. Alternatively we could rename all fields to the "start/end" convention to make RnaFusion consistent with LinxFusion

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/linx/LinxFusion.java
```diff
@@ -7,11 +7,32 @@
...
+    String geneContextEnd();
+
+    @NotNull
+    String geneTranscriptEnd();
+
+    @NotNull
+    default String name()
+    {
+        return String.format("%s-%s", geneStart(), geneEnd());
+    }
```
kduyvesteyn, line 35: I'd rename to `display` since this is effectively what it is. There should be no logic dependent on the actual implementation of this function and we have several other `display` functions in other places of the datamodels for similar purpose (mostly enums)

orange/src/main/java/com/hartwig/hmftools/orange/algo/isofox/FusionNameUtil.java
```diff
@@ -0,0 +1,25 @@
+package com.hartwig.hmftools.orange.algo.isofox;
+
+import com.hartwig.hmftools.common.rna.RnaFusion;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FusionNameUtil
+{
```
kduyvesteyn, line 9: Class should be `final`

orange-datamodel/src/test/resources/minimally.empty.orange.json
```diff
@@ -122,6 +121,7 @@
     "purpleClonalityPlot": "",
     "purpleCopyNumberPlot": "",
     "purpleVariantCopyNumberPlot": "",
-    "purplePurityRangePlot": ""
+    "purplePurityRangePlot": "",
+    "purpleKataegisPlot": "plot/tumor_sample.somatic.rainfall.png"
```
kduyvesteyn, line 125: Why do we need to provide an actual path here while the other plots can be empty?

orange-datamodel/src/test/resources/minimally.populated.orange.json
```diff
@@ -558,6 +557,7 @@
     "purpleClonalityPlot": "",
     "purpleCopyNumberPlot": "",
     "purpleVariantCopyNumberPlot": "",
-    "purplePurityRangePlot": ""
+    "purplePurityRangePlot": "",
+    "purpleKataegisPlot": "plot/tumor_sample.somatic.rainfall.png"
```
kduyvesteyn, line 561: Why do we need to provide a value here while all other plots are empty?

kduyvesteyn (commented):
In addition to comments here: There is some room for simplification in OrangeAlgo, line 693 onwards (handling of kataegis plot can be done similar to other purple plots)

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/isofox/GeneExpression.java
```diff
@@ -8,18 +8,20 @@
...
     double percentileCancer();
 
-    double medianTpmCohort();
+    @Nullable
+    Double medianTpmCohort();
 
-    double percentileCohort();
+    @Nullable
+    Double percentileCohort();
 }
```
kduyvesteyn, line 27: Apologies, just realized I mixed up these 2 fields. The `Cancer` fields need to be nullable rather the cohort fields. The cohort fields are "with respect to entire database" and hence are never null. The cancer fields are "with respect to cancer type" which can be null in case specific sample is missing cancer type (or unknown)

orange/src/main/java/com/hartwig/hmftools/orange/conversion/IsofoxConversion.java
```diff
@@ -1,39 +1,48 @@
...
+
+    @NotNull
+    public static GeneExpression convert(@NotNull com.hartwig.hmftools.common.rna.GeneExpression geneExpression)
     {
         return ImmutableGeneExpression.builder()
-                .geneName(geneExpression.geneName())
+                .gene(geneExpression.geneName())
                 .tpm(geneExpression.tpm())
                 .medianTpmCancer(geneExpression.medianTpmCancer())
                 .percentileCancer(geneExpression.percentileCancer())
```
kduyvesteyn, line 48: We should explicitly convert magic value -1 to null here.

kduyvesteyn (commented):
A few final remarks:
 - `FusionNameUtil` is now `public final` rather than `final public` following standard conventions
 - Have renamed the input parameter `experimentDate` to `samplingDate` to avoid unnecessary differences between input and output of ORANGE
 - Have added a brief summary of upcoming changes in README

PR #471, opened 2023-10-09
ACTIN-336: Make ORANGE disclaimer more clear

kduyvesteyn:
See attached for new example report including disclaimer.

[Test.orange.pdf](https://github.com/hartwigmedical/hmftools/files/12848398/Test.orange.pdf)


kzuberihmf (approved):
Looks reasonable to me and the output appears correct.

PR #473, opened 2023-10-12
ACTIN-322: Implement safe handling of missing linx plot directory

kzuberihmf:
This makes the linx plot directory config setting Nullable. When its null, linx plots are ignored. If not null, we check how many plots are expected and compare with plots found, and only error if a mismatch is detected.

So now if no plots are produced, it's ok for the plots dir to be missing (can still be specified in the orange config). It should be possible to cleanup manual creation of this directory in pipeline scripts.

To compute expected counts, this uses the same algorithm as linx visualizer (not in common code unfortunately), so the results should be the same. 

Notes this adds a dependency on two additional files in the linx directory not previously read by orange:  sample.linx.vis_fusions.tsv and sample.linx.vis_sv_data.tsv. For simplification, the entire data files are not modelled as they are in linx visualizer, instead only minimal fields required for checking are loaded.

Seemed to hold up in some manual testing. However as I've never encountered actual error modes of the pipeline process, not sure if there's a way for this checking to still fail (e.g. by the vis files not existing).


orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -658,25 +659,45 @@ private Map<PercentileType, Evaluation> evaluateCohortPercentiles(@NotNull Orang
...
+            try
+            {
+                linxReportableClusters =
+                        LinxReportableClusters.findReportableClusters(linxData, config.linxSomaticDataDirectory(), config.tumorSampleId());
+            }
+            catch(IOException e)
+            {
+                throw new RuntimeException("Unable to determine linx reportable clusters", e);
+            }
+
```
kduyvesteyn, line 691: I would separate loading and evaluation. It's cleaner to add the new structures to LinxData. Even though that class exists in hmf-common, it is only used by ORANGE and already a bridge class holding all data from linx that is relevant for ORANGE.

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -658,25 +659,45 @@ private Map<PercentileType, Evaluation> evaluateCohortPercentiles(@NotNull Orang
...
             LOGGER.info(" Loaded {} linx plots from {}", linxDriverPlots.size(), linxPlotDir);
+            if(linxReportableClusters.size() != linxDriverPlots.size())
+            {
+                throw new RuntimeException(String.format("Expected %d Linx plots but found %d", linxReportableClusters.size(), linxDriverPlots.size()));
+            }
```
kduyvesteyn, line 696: Are you confident your logic is always correct? 

I'd also be fine with a WARN in some situation I think, but leave that decision up to you. 

E.g. if there are 2 reportable clusters but 3 plots, would we be fine with a WARN?

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClustersTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    public void shouldFindAllReportableClusters() throws IOException
+    {
+        LinxData linxData = testLinxData();
+        Set<Integer> clusters = LinxReportableClusters.findReportableClusters(linxData, LINX_TEST_DATA_DIR, "tumor_sample");
+        assertEquals(new HashSet<>(Set.of(10, 50, 100)), clusters);
+    }
+
+    private static LinxData testLinxData()
```
kduyvesteyn, line 34: Missing @NotNull annotation, and its more typical to call such methods something like `createTestLinxData()`

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
kzuberihmf: Ah, ok didn't realize LinxData was orange specific. That does make the particular data structures I chose (simple lists and maps of integer ids), seem untidy ... in the current implementation I can get it away with this since its private to a class with a simple public interface.

I'll start by moving the current structures to LinxData + LinxDataLoader as the common location does seem to make the data organization in orange nicer. Then can see if it looks like we'd want more conventional structures in LinxData.

kduyvesteyn (approved):

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
kduyvesteyn: I'm fine with how LinxData currently looks. I haven't always mapped entire datamodels to their java representation in the past either, as long as it is clear what is meant by the datamodel.

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
```diff
@@ -140,17 +164,91 @@ private static List<LinxBreakend> selectReportableBreakends(@NotNull List<LinxBr
...
+    @NotNull
+    private static Map<Integer, Integer> loadSvToCluster(@NotNull String filename) throws IOException
+    {
+        List<String> lines = Files.readAllLines(new File(filename).toPath());
+        if(lines.isEmpty())
+        {
+            throw new IllegalStateException(String.format("File lacks header: %s", filename));
+        }
+
+        Map<Integer, Integer> svTocluster = Maps.newHashMap();
```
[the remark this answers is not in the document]
kduyvesteyn: Also, I notice you occasionally use guava functions and occasionally you don't use them.

For ACTIN we are on a path for guava-out but for hmftools that's not the intention. 

I am fine either way (guava yes/no) but may be good to be aware!

PR #474, opened 2023-10-13
prototype of adding a level of argument grouping to ConfigBuilder

kzuberihmf:
**Not for merging**

This is a small prototype of adding argument grouping into our CLI support, intended for design review. A simple implementation turned out to be straightforward to add.

For ACTIN-273, we want to disallow invalid combinations of arguments to orange, which is complicated by the different modes of operation (targeted vs whole-genome etc). Many CLI parsers support sub-commands, this example adds a very simple subcommand argument grouping mechanism to ConfigBuilder. 

The unit test gives an example. Subcommands are optional - current  ConfigBuilder users not affected. When subcommands are configured, each gets its own set of argument specifications, and global level arguments will be shared across all sub-commands. When configured, the first argument to the application must be a valid subcommand. Multiple sub-commands in a single invocation are not allowed.

Something like the following:

```bash
# run myapp in cmd1 mode that requires -arg1 and -arg2. 
$ myapp cmd1 -arg1 val1 -arg2 val2 -global_arg global_val
# run myapp with cmd2 that requires -arg3.
$ myapp cmd2 -arg3 val1 -global_arg global_val
```

Does this seem like enough flexibility to cover the needs of ACTIN-273, while pushing argument consistency verification into ConfigBuilder and not having a bunch of custom verification in the application code?

### DIFF
--- a/hmf-common/src/main/java/com/hartwig/hmftools/common/utils/config/ConfigBuilder.java
+++ b/hmf-common/src/main/java/com/hartwig/hmftools/common/utils/config/ConfigBuilder.java
@@ -2,16 +2,17 @@ package com.hartwig.hmftools.common.utils.config;
 
 import static java.lang.String.format;
 
-import static com.hartwig.hmftools.common.utils.config.ConfigUtils.setLogLevel;
-import static com.hartwig.hmftools.common.utils.file.FileWriterUtils.checkAddDirSeparator;
 import static com.hartwig.hmftools.common.utils.config.ConfigItemType.DECIMAL;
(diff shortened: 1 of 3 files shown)


kzuberihmf:
An alternative to adding subcommands to the parser would be to have multiple entry points to orange, each with a set of CLI args tailored to a specific type of operation. This would have the advantage of not requiring a more complex argument parser. Thanks to @jbartletthmf for the suggestion.

kzuberihmf:
Thanks for the discussion, closing this PR.

PR #476, opened 2023-10-17
ACTIN-212: report multiple non-canonical variants on same driver gene

kzuberihmf:
Small update to pick out all variants instead of the first in the list, plus test case.

### DIFF
--- a/orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
+++ b/orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
@@ -134,4 +134,56 @@ public class VariantEntryFactoryTest
                         .codingEffect(PurpleCodingEffect.MISSENSE)
                         .build()));
     }
+
+    @Test
+    public void shouldReportMultipleVariantsOnDriverGene()
+    {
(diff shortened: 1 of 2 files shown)


orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
```diff
@@ -134,4 +134,56 @@ public void canDetermineTranscriptImpact()
                         .codingEffect(PurpleCodingEffect.MISSENSE)
                         .build()));
     }
+
+    @Test
+    public void shouldReportMultipleVariantsOnDriverGene()
```
kduyvesteyn, line 139: I would rename this function, `shouldGenerateMultipleVariantsForNonCanonicalAndCanonicalDriver`. 
The current name suggests this specific function relates to "reported", but that only holds for non-canonical drivers.  Also I would reflect that this function specifically checks canonical in combination with non-canonical drivers.

kduyvesteyn (approved):
Changes look good.

Did you verify the solution against the sample that is mentioned in the jira though?

kzuberihmf:
> Did you verify the solution against the sample that is mentioned in the jira though?

Good point. Ran the update, it now produces one additional driver variant in the Orange report, an additional alt for CDKN2A.

PR #477, opened 2023-10-18
ACTIN-273: Followup on potential ORANGE issues in v5.33 pipeline

kzuberihmf:
* Updates lilac tables to show NONE instead of 0 for ref/rna support when missing. 
* Updates kataegis plot with purple fix

Orange config changes have been dropped from this PR and will be handled in a separate issue.


orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -227,6 +227,12 @@ public OrangeRecord run(@NotNull OrangeConfig config) throws IOException
             LOGGER.info("Wild-type calling skipped due to insufficient tumor sample quality");
         }
 
+        // TODO update haveRef flag after LILAC v1.6 is released, which fixes tumor-only fragments
+        // being written to the ref field.
+        //   boolean haveRef = config.refSampleWGSMetricsFile() != null && config.refSampleFlagstatFile() != null;
+        boolean haveRef = true;
+        boolean haveRna = config.rnaConfig() != null;
```
kduyvesteyn, line 234: Minor, but isn't proper English `hasRna` and `hasRef`?

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/HLAAlleleTable.java
```diff
@@ -51,14 +52,26 @@ public static Table build(@NotNull String title, float width, @NotNull List<Lila
...
     }
 
     @NotNull
-    private static String fragmentString(int fragments, boolean isTumorFail)
+    private static String fragmentString(@Nullable Integer fragments, boolean isTumorFail)
     {
-        return !isTumorFail ? String.valueOf(fragments) : ReportResources.NOT_AVAILABLE;
+        if(fragments == null)
+        {
+            return ReportResources.NONE;
```
kduyvesteyn, line 65: I'd think NOT_AVAILABLE would be better than NONE? 

I am fine if you want to stick to NONE but that term sounds a bit too similar to 0 which is also wrong.

kduyvesteyn (approved):
Approved pending some minor comments / thoughts

PR #478, opened 2023-10-19
ACTIN-282: Fix notation (un)disrupted copy numbers in ORANGE

cbruel:
In case of a hom_dup_disruption show undisrupted CN of a DUP as undisrupted CN minus junction CN


kduyvesteyn (approved):
Approved with some minor comments.

One general remark: Some files have formatting issues. If you auto-reformat your files this can't happen, but it seems that hasn't happened!

kduyvesteyn (approved):
Looks impeccable (means "onberispelijk" in Dutch!)

PR #480, opened 2023-10-19
ACTIN-341: Write exonUp and exonDown to database

kduyvesteyn:
See ticket for rationale.

Some additional notes:
 - This PR fixes an issue with writing `nextSpliceExonRank` which is currently a tinyint (max 255) which is too small for breakends in TTN (>300 exons). This PR changes this field to a small int.
 - For the patch I have decided to drop and recreate the database, since the new fields `exonUp` and `exonDown` are not null and have no meaningful default


kzuberihmf (approved):
Not super familiar with the db schema, but changes look ok to me. Ran the generate_database and migration patch script locally and ran cleanly. 

Though I notice that I need to change my db charset/collation because mysql default doesn't allow indices on the VARCHAR(1500) fields that we have. But that seems like a db admin issue and not necessarily a problem with our schema.

PR #481, opened 2023-10-19
ACTIN-352: Revise orange config to clearly separate modes of operation

kzuberihmf:
Adds experiment-type cli argument, and slits WGS config to a separate class that contains settings for both WGS tumor only and WGS tumor ref modes.


orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -115,10 +92,9 @@ public interface OrangeConfig
 
     static void registerConfig(@NotNull ConfigBuilder configBuilder)
     {
+        configBuilder.addConfigItem(EXPERIMENT_TYPE, true, "The type of the experiment, one of WGS or PANEL");
+        configBuilder.addConfigItem(TUMOR_SAMPLE_ID, true, "The sample ID for which ORANGE will run.");
         configBuilder.addConfigItem(TUMOR_SAMPLE_ID, true, "The sample ID for which ORANGE will run.");
```
kduyvesteyn, line 97: TUMOR_SAMPLE is added twice

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -155,25 +129,21 @@ static void registerConfig(@NotNull ConfigBuilder configBuilder)
         configBuilder.addPath(LINX_PLOT_DIR_CFG, false, LINX_PLOT_DIR_DESC);
         configBuilder.addPath(LINX_GERMLINE_DIR_CFG, false, LINX_GERMLINE_DIR_DESC);
```
kduyvesteyn, line 130: Linx germline should be registered by `OrangeWgsConfig`?

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
kduyvesteyn, line 1: SAGE_GERMLINE config registration should move to OrangeWgsConfig?

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -253,39 +208,18 @@ static void registerConfig(@NotNull ConfigBuilder configBuilder)
...
-    String sigsAllocationTsv();
-
     boolean convertGermlineToSomatic();
 
     boolean limitJsonOutput();
 
     boolean addDisclaimer();
 
+    @Nullable
+    OrangeWgsConfig wgsConfig();
```
kduyvesteyn, line 218: I think I would put this at the top, together with the Rna config

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -253,39 +208,18 @@ static void registerConfig(@NotNull ConfigBuilder configBuilder)
...
+
     default boolean tumorOnlyMode()
     {
-        return referenceSampleId() == null || referenceSampleId().isEmpty();
+        return wgsConfig() == null || wgsConfig().referenceSampleId() == null || wgsConfig().referenceSampleId().isEmpty();
```
kduyvesteyn, line 222: empty reference sampleId should not be a condition here, that's not really a valid value anyways (it either has to be present or null.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeWgsConfig.java
```diff
@@ -0,0 +1,193 @@
...
+import com.hartwig.hmftools.common.utils.config.ConfigBuilder;
+import com.hartwig.hmftools.common.virus.AnnotatedVirusFile;
+
+import org.immutables.value.Value;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@Value.Immutable
+@Value.Style(passAnnotations = { NotNull.class, Nullable.class })
+public interface OrangeWgsConfig
```
kduyvesteyn, line 49: The term `Wgs` as well as `WGS` exist within hmftools. In ORANGE, this concept is always called `WGS` so far so might be good to remain consistent? Also I think I would add "Ref" in the name somewhere, to make it clearer this is the WGSRefConfig!

orange/src/main/java/com/hartwig/hmftools/orange/OrangeWgsConfig.java
```diff
@@ -0,0 +1,193 @@
...
+
+        String cuppaDir = getToolDirectory(configBuilder, pipelineSampleRootDir, sampleDataDir, CUPPA_DIR_CFG, CUPPA_DIR);
+        if(cuppaDir == null)
+        {
+            throw new IllegalArgumentException("Cuppa dir not found but required for WGS config");
+        }
+
+        builder.cuppaResultCsv(fileIfExists(CuppaDataFile.generateFilename(cuppaDir, tumorSampleId)));
+        builder.cuppaSummaryPlot(fileIfExists(CuppaDataFile.generateReportSummaryPlotFilename(cuppaDir, tumorSampleId)));
+        builder.cuppaFeaturePlot(optionalFileIfExists(CuppaDataFile.generateReportFeaturesPlotFilename(cuppaDir, tumorSampleId)));
```
kduyvesteyn, line 145: This is correct, but one note for clarification: Cuppa either fits on one page or on two pages. If Cuppa requires 2 pages it generates a second file called 'feature plot'. In case CUPPA fits on one page, the feature plot is included in the summary plot. That's why the feature plot is truly optional (like linx plot dir)

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -525,12 +523,14 @@ private static LilacSummaryData loadLilacData(@NotNull OrangeConfig config) thro
     @Nullable
     private static VirusInterpreterData loadVirusInterpreterData(@NotNull OrangeConfig config) throws IOException
     {
+        // TODO consider if the tumor only mode is better replaced with checking if WGS is configured
+        //  here and elsewhere below
```
kduyvesteyn, line 527: This is definately better. Best to rely on the source rather than some derived value. If there is a virus dir present, we can assume virus data can be loaded, unrelated to modes or ref/no-ref etc

kzuberihmf: I think my TODO comment was already made obsolete by having changed the tumor only check to consider wgsRefConfig.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeWgsConfig.java
kzuberihmf: Ok. Note we did end up maintaining the WGS tumor only possibility, at least to the level it was previously supported which may have been just partial.

kduyvesteyn (approved):
Approved with 2 potential remaining suggestions:
 - I would suggest to remove the property `OrangeConfig.tumorOnlyMode`. Its usages are only in OrangeAlgo and usage is not consistent. I would suggest to strictly use the direct variables, e.g. for `loadSigAllocations` you should simply load in case wgsRefConfig != null.
 - The PathResolver (and Config) class would benefit from unit testing I'd think. But can be left for future since OrangeAlgoTest does cover these functions pretty well.

PR #487, opened 2023-11-03
ACTIN-376: support verifier use multiple data model jars by shading

kzuberihmf:
Adds a profile to build a shaded jar for orange-datamodel.


pauldwolfe (approved):
I guess to release this we'll need to make it datamodel version `2.1.0`?

PR #496, opened 2023-12-14
AUS-204: CUPPAv2 output into ORANGE; Moved ORANGE changes from master to AUS-204 branch

luan-n-nguyen:
Hi all,

Previously the commits related to ORANGE were on master (which they shouldn't have been). I've reverted these commits on the master branch and cherry picked them on the AUS-204 branch to move them there.

You can ignore commits related to the cuppa/src/main/python/pycuppa/ directory which aren't related to ORANGE. I just had to move this dir out of cuppa/src/main/resources so that the cuppa module builds successfully.

Luan


kzuberihmf (commented):
Hi Luan, just want to double check what's in scope for this update. For example this PR seems to show just a small change in orange-datamodel to cuppa2/Cuppa2Data.java, but if I understand the intention then all of the files under that cuppa2 folder should be part of this change? 

In addition, I see that the orange record now includes both cuppa and cuppa2 (I assume so as not to break compatibility for users of orange-datamodel). Is the intention to only support cuppa2 going forward or both in some type of either-one-or-the-other setup? For example, I think there are some parts of the original cuppa record still being used by orange in its report generation, however I'm not sure it is loaded anymore by orange algo?

Apologies if I'm misunderstanding, just want to get a handle on the big picture, will be looking at the changes in more detail.

kzuberihmf:
Since the diff view isn't showing what I think are the relevant files, I'm going to try just linking to lines in the current head of this branch [39827e4](https://github.com/hartwigmedical/hmftools/commit/39827e4433c78190334450e79146f275b879c536) in comments. Hope this proves useful.

In addition, I'll note up front that the coding style for orange & orange-datamodel is somewhat particular (I've been learning this the hard way 😄), emphasizing a strong internal consistency. This includes for example preference for clarity over brevity of identifiers, minimizing redundancy, avoiding dead/comment-out code, annotating nullability for all non-primitive types in function signatures and attribute declarations and so on.

As with all such guidelines, I think exceptions may be acceptable where justified. And I'm not super-expert at it myself, but let's have a go! 

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/Cuppa2Data.java#L19

How about expanding to probabilities()

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/Cuppa2Data.java#L22

featureContributions()?

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/Cuppa2Data.java#L22

signatureQuantiles()?

And backing up a bit:

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/Cuppa2Data.java#L16

I guess this is a convenience attribute for an entry that's already in probs(). Do you think its might be more useful in general to instead have an ordering helper? E.g. we have helpers elsewhere in orange-datamodel to compute derivable attributes such as for fusion names, though the precedent might not be super-strong yet. Also is the ordering by probability with the rank & rankgroup pre-computed from this? What happens if there is a tie for the top prediction, only one can be represented.

Continuing with ProbabilityEntry:

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/ProbabilityEntry.java#L13

Not clear what this means, is it a standard acronym everyone knows or better understood if expanded? Also, is it NotNull?

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/ProbabilityEntry.java#L14

NotNull?

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/ProbabilityEntry.java#L15

Pretty generic term, but is this the probability, or what Cuppa-1 calls likelihood? Is it different enough that we should change its name?

BTW, could you make the layout consistent with the other data classes in datamodel, it looks like we prefer a single empty line between each attribute entry.

kzuberihmf:
Continuing on:

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/FeatureContributionEntry.java#L14-L15

feat -> feature. Also for the rest of FeatureContributionEntry the comments from earlier apply, nullability annotations, layout, and attribute naming as mentioned before (whichever names we we settle on for best clarity).

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/cuppa2/SignatureQuantileEntry.java#L13-L18

As above.

One more on naming:

https://github.com/hartwigmedical/hmftools/blob/39827e4433c78190334450e79146f275b879c536/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangePlots.java#L45-L55

Only the last entry is cuppa2. Looking at the names, cuppaSummaryPlot and cuppaFeaturePlot are quite clear. cuppaChartPlot not so much (not part of this change its something we already had), it reads to me like 'cuppaPlotPlot' 😄. For cuppa2VisPlot I see we're being consistent with naming elsewhere e.g. CuppaVisData in hmf-common, though as orange-datamodel is an interface to external users I don't think we are bound to the internal naming. So if there's a nice clear plot name it could be nice here.

I can't be certain from the incomplete diff if orange-datamodel cuppa2 fields are covered in the unit tests, but searching the source I guess not yet. Wondering how we could generate a datamodel file that includes cuppa2 fields 🤔 to test against, as we do with assertCuppa() in OrangeJsonTest.

kzuberihmf:
The changes in this PR to amber, purple and a bunch of readme's etc are not logically part of this change right? I'm not familiar with those and won't review. Also there's cuppa changes to patient-db that I'm not familiar with and not really looking at.

BTW I can't build the complete hmftools repo in this branch, something to do with cider and liftover api usage. But cuppa and orange build so I'm assuming that's irrelevant, just mentioning in case its part of a mis-merge or something that requires action.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeWGSRefConfig.java
```diff
@@ -76,17 +75,14 @@ static void registerConfig(@NotNull ConfigBuilder configBuilder)
...
+    @Nullable
+    String cuppa2VisData();
 
     @Nullable
-    String cuppaFeaturePlot();
+    String cuppa2VisPlot();
 
     @Nullable
-    String cuppaChartPlot();
+    String cuppa2PredSumm();
```
kzuberihmf, line 85: Prediction summary? Also I didn't notice where this was used, other than being populated.

Also, these three cuppa2 entries are all paths, e.g. elsewhere you used `String cuppaVisDataTsv = config.wgsRefConfig().cuppa2VisData();` so maybe it makes sense to just name the field with the Tsv suffix as well? We also seem to use the suffixes File, Directory where appropriate. But plot files are just called ...Plot for whatever reason, so that one still works 😄.

kzuberihmf: Also, I believe @kduyvesteyn was interested in a way for the cuppa run mode to be passed through to orange. Didn't notice that that was possible yet, can it be unambiguously determined from the cuppa2 data that is loaded already, or do we need an additional param to be passed in by the caller of orange for this to be possible?

kzuberihmf:
Could you update the orange reamde section `Version History and Download Links` for this change? Also orange & orange-datamodel will need minor version bumps, though maybe that can be taken care of in master as part of the release procedure.

kzuberihmf:
Thanks for the PR Luan. I wasn't perfectly clear on the requirements, but from reading I understand that the idea is that cuppa2 is implemented into orange as a separate data record from the original cuppa1, as augmenting cuppa1 with additional attributes to accommodate both wasn't feasible. The cuppa1 record continues to exist in orange-datamodel but orange from now will not populate it. 

I haven't manually verified, but existing orange-datamodel users should be able to use the class to load json from the previous datamodel without breaking, but will have to explicitly check the two cuppa fields if they need to support both versions for some reason. In other words, future users of orange-datamodel written for cuppa2 will not be able to just load a cuppa1 orange-datamodel json and just happily get a subset of the data, rather they will not see cuppa at all unless they explicitly code for it. I assume this was agreed on in the design but don't actually know for sure so mentioning here to be super-extra explicit 😆.

A concern I have is that some attributes from cuppa1 continue to be used in orange, for example CuppaData.maxComplexSize() is reported by orange on the report (FrontPageChapter.java). All such fields will now be reported as NOT_AVAILABLE, until we either remove or replace them with equivalent. Not sure what the path forward is for this, but I think it needs to be sorted out before the next orange release.

luan-n-nguyen:
> Hi Luan, just want to double check what's in scope for this update. For example this PR seems to show just a small change in orange-datamodel to cuppa2/Cuppa2Data.java, but if I understand the intention then all of the files under that cuppa2 folder should be part of this change?
> 
> In addition, I see that the orange record now includes both cuppa and cuppa2 (I assume so as not to break compatibility for users of orange-datamodel). Is the intention to only support cuppa2 going forward or both in some type of either-one-or-the-other setup? For example, I think there are some parts of the original cuppa record still being used by orange in its report generation, however I'm not sure it is loaded anymore by orange algo?
> 
> Apologies if I'm misunderstanding, just want to get a handle on the big picture, will be looking at the changes in more detail.

Hi Khalid,

Thanks for the detailed review! Paul requested to have the CUPPA v1 fields in orange-datamodel. I've kept as much CUPPA v1 code as i could in orange, though this wasn't a specific request from Paul. @pauldwolfe Shall i leave the  CUPPA v1 code in orange or shall i remove it completely?

orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa2/Cuppa2DataFactory.java
```diff
@@ -0,0 +1,113 @@
...
+    @NotNull
+    public static ImmutableProbabilityEntry getTopPrediction(CuppaVisData visData)
+    {
+        Categories.ClfName mainCombinedClfName = visData.getMainCombinedClfName();
+
+        CuppaVisDataEntry visDataEntry = visData
+                .subsetByDataType(Categories.DataType.PROB)
+                .subsetByClfName(mainCombinedClfName)
+                .getTopPredictions(1)
+                .get(0);
```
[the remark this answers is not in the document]
luan-n-nguyen: > Can tied-ranks happen?

Ties cannot happen. The rank attribute is precomputed in python (disallowing ties) and written to the tsv that's loaded into orange

orange/src/main/java/com/hartwig/hmftools/orange/OrangeWGSRefConfig.java
luan-n-nguyen: I've now removed the unused`cuppa2PredSumm()` property

orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa2/Cuppa2DataFactory.java
```diff
@@ -0,0 +1,113 @@
...
+        return ImmutableCuppa2Data.builder()
+                .topPrediction(getTopPrediction(visData))
+                .probs(getProbabilities(visData))
+                .featContribs(getFeatureContributions(visData))
+                .sigQuantiles(getSignatureQuantiles(visData))
+                .build();
+    }
+
+    @NotNull
+    public static ImmutableProbabilityEntry getTopPrediction(CuppaVisData visData)
```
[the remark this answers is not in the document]
luan-n-nguyen: I've now added NotNull/Nullable annotations to ProbabilityEntry, FeatureContributionEntry, and SignatureQuantileEntry classes

kzuberihmf:
Thanks Luan. Would you happen to know if attributes equivalent the cuppa1 maxComplexSize, telomericSGLs and LINECount exist somewhere in the cuppa2 output?

pauldwolfe:
> > Hi Luan, just want to double check what's in scope for this update. For example this PR seems to show just a small change in orange-datamodel to cuppa2/Cuppa2Data.java, but if I understand the intention then all of the files under that cuppa2 folder should be part of this change?
> > In addition, I see that the orange record now includes both cuppa and cuppa2 (I assume so as not to break compatibility for users of orange-datamodel). Is the intention to only support cuppa2 going forward or both in some type of either-one-or-the-other setup? For example, I think there are some parts of the original cuppa record still being used by orange in its report generation, however I'm not sure it is loaded anymore by orange algo?
> > Apologies if I'm misunderstanding, just want to get a handle on the big picture, will be looking at the changes in more detail.
> 
> Hi Khalid,
> 
> Thanks for the detailed review! Paul requested to have the CUPPA v1 fields in orange-datamodel. I've kept as much CUPPA v1 code as i could in orange, though this wasn't a specific request from Paul. @pauldwolfe Shall i leave the CUPPA v1 code in orange or shall i remove it completely?
> 
> In any case, I've incorporated some of your suggested changes (mainly renaming of attributes). I'll work your other suggestions in the coming days.

Hey @luan-n-nguyen , I asked that we maintain backwards compatibility in the orange datamodel. It d

> > Hi Luan, just want to double check what's in scope for this update. For example this PR seems to show just a small change in orange-datamodel to cuppa2/Cuppa2Data.java, but if I understand the intention then all of the files under that cuppa2 folder should be part of this change?
> > In addition, I see that the orange record now includes both cuppa and cuppa2 (I assume so as not to break compatibility for users of orange-datamodel). Is the intention to only support cuppa2 going forward or both in some type of either-one-or-the-other setup? For example, I think there are some parts of the original cuppa record still being used by orange in its report generation, however I'm not sure it is loaded anymore by orange algo?
> > Apologies if I'm misunderstanding, just want to get a handle on the big picture, will be looking at the changes in more detail.
> 
> Hi Khalid,
> 
> Thanks for the detailed review! Paul requested to have the CUPPA v1 fields in orange-datamodel. I've kept as much CUPPA v1 code as i could in orange, though this wasn't a specific request from Paul. @pauldwolfe Shall i leave the CUPPA v1 code in orange or shall i remove it completely?
> 
> In any case, I've incorporated some of your suggested changes (mainly renaming of attributes). I'll work your other suggestions in the coming days.

Hey both,

The request I made was that we should make every effort to make orange backwards compatible to its downstream clients (actin, oncoact reporting, etc). As @kzuberihmf mentions, that means both maintaining the structure, but also ensuring those values are populated. I didn't intend to maintain separate datamodels for the different versions, unless we were to somehow back-populate the old one with whatever relevant data from cuppa v2. 

To give a concrete example, instead of introducing the new Cuppa2Data +  XxxEntry classes, to be compatible you would expand the CuppaPrediction to include any new fields/metrics and still populate the old ones which are relevant. As all fields other than likelihood are nullable, you could maintain compatibility by maintaining the list of predictions in rank order and then add the new probabilities/features/sig quantiles alongside it. 

This might be challenging, impractical or maybe impossible as the orange datamodel exposes internal cuppa implementation details which perhaps are not portable between versions. Regardless, the current implementation adds additional complexity of a second model yet it does not provide any backward compatibility, so we'll need to rethink it.  

Apologies for not being more clear on what I meant previously. 

An alternative to all this would be to break backwards compatibility. As I learn more about the details of the orange datamodel, I'm realizing that keeping it backwards compatible in its current form might be very challenging when there are significant changes in tools. Maybe a new major version of a tool is a big enough of an event to always warrant a compatibility break. There is cost of this is coordination between applications, but it's not that often we roll out big changes to existing applications such as this. 

So I'd suggest we do the following:
- See if we can find a way to rescue this PR from git hell.  In it's current state it's not possible to see what's actually changed in Orange and given the size of the change I think reviewing it in place in the code will be risky.
- Have a go at making it truly backwards compatible using a shared model expanded for new cuppa 2 features. I'm fine to time-box this to a couple of days, and to concede to @luan-n-nguyen if he feels its not possible.
- Go for a non-backwards compatible change to support cuppa 2 and make this orange datamodel v3. 

@luan-n-nguyen realize that you may still have some questions and if might be good to have a chat before investing more time. Let me know via teams if you want to do a call tomorrow morning Sydney time, I can log on in the evening.

kzuberihmf:
Hi @luan-n-nguyen, looks like this branch was superseded by AUS-213 which was merged. If there's nothing left outstanding here, you can probably just close this PR?

luan-n-nguyen:
> Hi @luan-n-nguyen, looks like this branch was superseded by AUS-213 which was merged. If there's nothing left outstanding here, you can probably just close this PR?

Done!

PR #497, opened 2023-12-21
ACTIN-46: Make health checker component for WTS

kzuberihmf:
This is a prototype reproducing the sql query logic we currently use. Its currently in patientdb simply because that's where I found the data loading and was interested in learning how our vcf parsing/processing works.

Before carrying any further, would like some clarification on if this is useful in general and belongs in hmftools. If so we could create a new module e.g. 'crest' as was suggested and move it there, as well as arrange for it to be packaged as a maven artifact for use in pipelines. 

Alternatively we could move it to another repo, but that means reproducing some of the supporting vcf/variant code that's in hmf-common as that doesn't seem to be packaged in maven yet.

Also comments if its generally on the right track welcome! Basically it just streams through the vcf and counts up the required totals, takes hardly a second to run. Of course still to do is proper arg parsing, output and error return codes etc.


kzuberihmf:
Ready for review. 

Using the purple dir config makes this a bit awkward to use directly after sage, as you have to rename and move the vcf around. Might be simpler to take the vcf file directly. Anyhow I followed what I assume is the usual pattern as in health-checker. And would like to get this wrapped up soon.

crest/src/test/java/com/hartwig/hmftools/crest/CrestTest.java
kduyvesteyn, line 1: You should omit the testing of the actual creation of the test files inside this code I'd think. Not everyone has "hmf/tmp", and build servers don't have it for sure. 

One alternative (e.g. done in ORANGE) is to have a separate TestMain (see `ReportGeneratorTestApplication` which is not a junit test but simply a main that we can run manually if we are in doubt of whether the succeeded/failed. Do note that the actual logic should be tested obviously, so maybe the function to evaluate should be separated from the function to write?

I realize there is also a reference to hmf/tmp in the health check application test but this should be removed for the same reason.

crest/src/test/resources/purple/tumor_sample.purple.germline.vcf.gz
kduyvesteyn, line 1: I typically try to use unzipped files as test files (see orange test_run), but in ORANGE there is specific handling for both zipped and unzipped, so that's not super clean either. 

In any case no strong opinion but at least would consider making this file unzipped for easier inspection / changing.

kduyvesteyn (commented):
Logic looks good! Number of comments on structure and tests. 

Also:
 - Have edited the example purple file to use "ref_sample", "tumor_sample", "rna_sample" as sample IDs, these are the typical IDs used for testing (no need to refer to a "magic" value such as COLO829v003T). One question: where does the test data actually come from?
 -  Have merged master into ACTIN-46 branch since ACTIN-46 had a failing unit test in NEO. 
 - Have adjusted some formatting myself rather than add in comments (see my commits)

crest/src/main/java/com/hartwig/hmftools/crest/CrestApplication.java
```diff
@@ -0,0 +1,205 @@
...
+            if(!sampleInFile(sampleToCheck, header))
+            {
+                throw new RuntimeException("Sample " + sampleToCheck + " not found in file " + vcfFilename);
+            }
+
+            for(VariantContext context : reader.iterator())
+            {
+                VariantContextDecorator decorator = new VariantContextDecorator(context);
+
+                if(decorator.filter().equals("PASS") && decorator.type() == VariantType.SNP && !decorator.gene().isEmpty())
```
kduyvesteyn, line 181: Cleaner to reuse `isPass` function on `VariantContextDecorator`?

crest/src/main/java/com/hartwig/hmftools/crest/CrestApplication.java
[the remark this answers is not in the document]
kzuberihmf: Test data is probably not ideal due to wanting colo wts, its something we had cooked up earlier for dev work, @pauldwolfe would know the backstory. More on test data below.

Thanks for just going ahead with formatting updates etc.

crest/src/test/java/com/hartwig/hmftools/crest/CrestTest.java
kzuberihmf: Good point about hmf/tmp not necessarily existing (yeah I stole it from health-check :smile:). Interestingly this was the one test that actually turned up a bug for me! In previous experience we've sometimes used temp dir creation functionality that exists in a test harness, it ends up feels more like an integration test than unit tests but does tend to turn up problems.

Anyhow, I'll think about refactoring/testing the logic more directly.

crest/src/test/resources/purple/tumor_sample.purple.germline.vcf.gz
kzuberihmf: Couple of issues here, one is that if I use `PurpleCommon.purpleGermlineVcfFile(purpleDir, sampleId)` it seems to impose the gz extension as well as applying our file naming conventions. As I mentioned I also find it awkward to impose that naming, do you think this could be less tied to our workflow layout and just take the full path/name of the vcf as input? Or I'm using the wrong api probably. 

Second is that I found the vcf is too noisy to be very helpful for manual inspection (long lines with lots of embedded fields). I agree with your perspective, and had actually started to edit down the vcf by removing records and INFO fields, realized that was annoying to do manually, and then stopped just before attempting to write a script to create a custom vcf 😄 (got too lazy). But actually what I want is a small input test file where I know what the results of the calculation should be. Here I'm actually relying on the fact that I verified manually against another other process on other data, and just am using using colo here to catch future regressions. So not ideal, I might revisit. A small custom vcf would also remove our having to wonder about where the test file came from.

crest/src/main/java/com/hartwig/hmftools/crest/CrestApplication.java
kzuberihmf: Didn't notice that one, thanks!

crest/src/test/resources/purple/tumor_sample.purple.germline.vcf.gz
kduyvesteyn: Some thoughts:
 - I would keep `purple_dir` as the top-level input, and resolve the VCF path from there. Your algo could take an actual path though, and that will allow you to test with an unzipped VCF. This would be the same structure as ORANGE -> OrangeConfig has actual paths (e.g. somatic vcf) as properties but in its factory takes paths to algo-output (e.g. purple)
 - You could have a "minimal" and a "real" vcf to test on. Both have their merits. Again, similar to ORANGE, where the test file "purple/variants.vcf" is a simple test file for unit testing, while "test_run" is the actual COLO output and serves more of an integration-test functionality (and this input is only used to assess ORANGE actually runs on real data, there is no data-checking done).

crest/src/main/java/com/hartwig/hmftools/crest/CrestAlgo.java
```diff
@@ -0,0 +1,158 @@
...
+public class CrestAlgo
+{
+    private static final Logger LOGGER = LogManager.getLogger(CrestAlgo.class);
+
+    @NotNull
+    private final String purpleDir;
+    @Nullable
+    private final String outputDir;
+    @NotNull
+    final String sampleId;
```
kduyvesteyn, line 36: Property can be private

crest/src/main/java/com/hartwig/hmftools/crest/CrestApplication.java
kduyvesteyn, line 1: FYI: I flipped the functions around such that the main function is on top

crest/src/test/java/com/hartwig/hmftools/crest/CrestTest.java
kduyvesteyn, line 1: Should be renamed to `CrestAlgoTest`

crest/src/test/resources/purple/tumor_sample.purple.germline.vcf.gz
kduyvesteyn, line 1: I suggest to replace this with a "dummy" VCF since you no longer functionally test using this data. Now you have a VCF with 1500 lines suggesting some complicated test is going on based on it.

kduyvesteyn (approved):
Looks good to me.

Related thought: This logic actually doens't depend on RNA strictly. The only thing this code those is that there is sufficient support within `sampleToCheck` for the variants found for `sampleId`. I wouldn't suggest to change anything but interesting to realize nonetheless :)

PR #502, opened 2024-01-09
DEV-3738 "At least" versions of cuppa-chart deps

nedleitch:
This should continue to work for users under older operating systems but also allow newer installations to use later versions of the described dependencies.


luan-n-nguyen:
Hi Ned,

I'm working (as we speak) on getting CUPPA v2 into pipeline 5.34. Should be done by the end of the week, if not sometime next week. I would say merge the PR for now since `cuppa-chart` will be removed very soon.

Luan

nedleitch:
Dropping this as we've already incorporated the new 2.0 Cuppa version in pipeline5 5.34.

PR #503, opened 2024-01-10
DEV-3727: Add option to PAVE to never PON-filter ClinVar pathogenic variants

DFKoetsier:
Intended for targeted runs. Applies to variants whose ClinVar annotations would be categorized as PATHOGENIC or LIKELY_PATHOGENIC by the code for such interpretations in hmf-common. Making correct use of this feature requires providing ClinVar and blacklist annotation in PAVE for the run.


DFKoetsier:
Charles agreed with merging this

PR #504, opened 2024-01-15
ACTIN-361: linx plot counts shouldn't include simple clusters

kzuberihmf:
Skips simple clusters when counting expected linx plots (reportable clusters):

* loads exon counts from linx.vis_gene_exon.tsv
* loads chain counts for each cluster id from linx.vis_sv_data.tsv
* Filters the count of reportable clusters to remove single link clusters that don't touch an exon, consistent with [this](https://github.com/hartwigmedical/hmftools/blob/7213faacc41f107a027bcf368ab6a094eb09ecd1/linx/src/main/java/com/hartwig/hmftools/linx/visualiser/SvVisualiser.java#L215).
 
Recall if the accounting does not match we get a warning in the orange output, this should fix a couple of rare cases.


orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClusters.java
```diff
@@ -30,6 +30,14 @@ public static Set<Integer> findReportableClusters(@NotNull LinxData linx)
...
         linx.somaticDrivers().stream().filter(x -> x.clusterId() >= 0).forEach(x -> clusterIds.add(x.clusterId()));
-        return clusterIds;
+
+        return clusterIds.stream().filter(x -> !isSimpleCluster(linx, x)).collect(toSet());
+    }
+
+    private static boolean isSimpleCluster(@NotNull LinxData linx, int clusterId)
+    {
+        boolean hasSingleLink = !linx.clusterIdToChainCount().containsKey(clusterId) || linx.clusterIdToChainCount().get(clusterId) == 1;
+        boolean hasNoExons = !linx.clusterIdToExonCount().containsKey(clusterId) || linx.clusterIdToExonCount().get(clusterId) == 0;
```
kduyvesteyn, line 40: Based on the actual gene exon vis data, it seems that these maps are sparse (not all clusters are present in the maps). 

Is it still correct to assume that if a cluster does not exist in the mappings, it is a simple cluster?

orange/src/test/resources/test_run/linx/tumor_sample.linx.vis_gene_exon.tsv
```diff
@@ -0,0 +1,23 @@
...
+COLO829v003T	65	PTEN	ENST00000371953	10	DRIVER	9	89725044	89731687
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	1	89622870	89624305
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	2	89653782	89653866
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	3	89685270	89685314
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	4	89690803	89690846
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	5	89692770	89693008
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	6	89711875	89712016
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	7	89717610	89717776
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	8	89720651	89720875
+COLO829v003T	68	PTEN	ENST00000371953	10	DRIVER	9	89725044	89731687
```
kduyvesteyn, line 23: References to COLO in every test file are supposed to be renamed to `tumor_sample` which is the sample ID used in the test run

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
```diff
@@ -121,7 +125,9 @@ private static LinxData load(@NotNull String somaticStructuralVariantTsv, @NotNu
                 .reportableSomaticBreakends(reportableSomaticBreakends)
                 .somaticHomozygousDisruptions(somaticHomozygousDisruptions)
                 .fusionClusterIds(fusionClusterIds)
-                .putAllSvIdToClusterId(svIdToClusterId)
+                .putAllSvIdToClusterId(visSvData.svIdToClusterId)
+                .putAllClusterIdToChainCount(visSvData.clusterIdToChainCount)
```
kduyvesteyn, line 129: Minor and up for debate, but I'd use getters for these (and make the properties private in VisSvData)

Another minor, maybe just set the properties in the builder instead of using `putAll`? To me, `putAll` gives the impression that is more is coming (you are putting entries into a growing map)

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
```diff
@@ -217,7 +223,7 @@ private static String generateVisSvDataFilename(@NotNull String basePath, @NotNu
     }
 
     @NotNull
-    private static Map<Integer, Integer> loadSvToCluster(@NotNull String filename) throws IOException
+    private static VisSvData loadSvToCluster(@NotNull String filename) throws IOException
```
kduyvesteyn, line 226: Minor but I'd rename the input variable to `somaticVisSvDataTsv` since you only support reading that specific file, and think it makes the code more readable now that multiple "vis files" are read

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
```diff
@@ -240,8 +247,69 @@ private static Map<Integer, Integer> loadSvToCluster(@NotNull String filename) t
             {
                 svToCluster.put(svId, clusterId);
             }
+
+            if(clusterIdToChainCount.containsKey(clusterId))
+            {
+                clusterIdToChainCount.put(clusterId, clusterIdToChainCount.get(clusterId) + 1);
```
kduyvesteyn, line 253: To double-check: Every line in the SV data for the same cluster represents an extra chain link?

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClusters.java
```diff
@@ -30,6 +30,14 @@ public static Set<Integer> findReportableClusters(@NotNull LinxData linx)
         svIds.stream().filter(svToCluster::containsKey).map(svToCluster::get).forEach(clusterIds::add);
 
         linx.somaticDrivers().stream().filter(x -> x.clusterId() >= 0).forEach(x -> clusterIds.add(x.clusterId()));
-        return clusterIds;
+
+        return clusterIds.stream().filter(x -> !isSimpleCluster(linx, x)).collect(toSet());
+    }
+
+    private static boolean isSimpleCluster(@NotNull LinxData linx, int clusterId)
```
kduyvesteyn, line 37: Minor but maybe rename to `isSimpleClusterAffectingNoExons`?

In my mind, a simple cluster is a cluster with no chains, but that is independent of whether exons are affected or not.

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClustersTest.java
```diff
@@ -24,13 +24,49 @@ public class LinxReportableClustersTest
...
+    @Test
+    public void shouldNotCountSimpleClusters()
+    {
+        // cluster 10 has single chain link and no exons -> should not be counted
+        LinxData linxData = linxDataBuilder()
+                .clusterIdToChainCount(Map.of(10, 1, 50, 2, 100, 2))
+                .clusterIdToExonCount(Map.of(50, 1, 100, 1))
+                .build();
+        Set<Integer> clusters = LinxReportableClusters.findReportableClusters(linxData);
+        assertEquals(new HashSet<>(Set.of(50, 100)), clusters);
```
kduyvesteyn, line 41: Is it necessary to create a HashSet from `Set.of`, doesn't that create something that is equal to `clusters` already?

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClustersTest.java
```diff
@@ -24,13 +24,49 @@ public class LinxReportableClustersTest
...
+    }
+
+    @Test
+    public void shouldCountMultiLinkClustersWithNoExon()
+    {
+        // cluster 10 has two chain links and no exons -> not simple so should be counted
+        LinxData linxData2 = linxDataBuilder()
```
kduyvesteyn, line 48: I wouldn't use "2" in the name, since the test is independent from 1 an 3. Either just `linxData` or a more descriptive name

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClusters.java
```diff
@@ -19,9 +20,8 @@ public final class LinxReportableClusters
     @NotNull
     public static Set<Integer> findReportableClusters(@NotNull LinxData linx)
```
kduyvesteyn, line 21: I think this name is no longer applicable. Simple clusters could be reportable but they are just not visualised? Maybe a better name would be something like `findVisualizedClusters`?

kduyvesteyn (approved):
Changes look good, added a bunch of minor comments.

Also: I assume change has been regression-tested on a bunch of samples and also on the samples where we expect a difference (see ACTIN-361 jira ticket)

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClustersTest.java
kzuberihmf: It does indeed!

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxReportableClustersTest.java
kzuberihmf: yeah I think I had these all in a single test before splitting up, and didn't renamed the vars.

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
kzuberihmf: That was my understanding, and my earlier testing seemed to work. SvVisualizer calls these records clusterLinks, my naming here is maybe not great.

hmf-common/src/main/java/com/hartwig/hmftools/common/linx/LinxDataLoader.java
```diff
@@ -303,13 +302,25 @@ private static Map<Integer, Integer> loadClusterExonCounts(@NotNull String filen
...
+    public Map<Integer, Integer> getSvIdToClusterId()
+    {
+        return svIdToClusterId;
+    }
+
+    @NotNull
+    public Map<Integer, Integer> getClusterIdToChainCount()
+    {
+        return clusterIdToChainCount;
+    }
```
kduyvesteyn, line 325: We typically don't use "get" for getters since that is implicit. These functions should be named `svIdToClusterId()` and `clusterIdToChainCount()`

kduyvesteyn (approved):
Looks good (one minor comment!)

PR #506, opened 2024-01-22
INFRA-204 make program exit if maven build fails

KasperWolsink99:
This should address the issue discovered in https://console.cloud.google.com/cloud-build/builds;region=europe-west4/c97c6129-b807-46df-852e-e42457b7d04d;step=0?project=hmf-build


nedleitch (approved):
I see some other `subprocess` calls in the script, seems pretty unlikely they would fail but I think it is possible since you're not using Maven at all so if it were me I'd add the check on those too. Nonetheless not going to hold this up, think it looks like it will fix the discovered issue.

PR #509, opened 2024-01-26
ACTIN-541: Throw exception in case isofox cancer type does not exist in gene distribution data

kduyvesteyn:


hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
```diff
@@ -109,4 +110,14 @@ public double getTpmPercentile(final String geneId, final String cancerType, fin
 
         return getPercentile(percentileMap.get(cancerType), sampleTpm);
     }
+
+    Set<String> configuredCancerTypes()
```
kzuberihmf, line 114: NotNull?

kzuberihmf (approved):

hmf-common/src/main/java/com/hartwig/hmftools/common/isofox/GeneExpressionDistributionData.java
kduyvesteyn: I considered it but there are no principles around this usage in hmftools so I usually keep the file I work in consistent (and this file didn't use nullable annotations)

PR #522, opened 2024-02-19
INFRA-266 Improve patient-db loaders

mdenburger:
Several small improvements and additions to patient-db that make it possible to run the loader classes as separate processes in an automated way:
* Let all loaders return exit code `1` on failure.
* Let all loaders use try-with-resources to close the database connection on failure.
* Add class `CheckSampleExistsInDatabase`, which is used by automation to check whether a sample already exists.
* Add a `-sample` parameter to `LoadCuppa2`. Using the TSV header doesn't work since that's often just "sample_1" instead of the real sample ID.

Also removed the "shade" profile, which is no longer needed when the `patient-db.jar` is run as a separate process.

### DIFF
--- a/patient-db/src/main/java/com/hartwig/hmftools/patientdb/CheckSampleExistsInDatabase.java
+++ b/patient-db/src/main/java/com/hartwig/hmftools/patientdb/CheckSampleExistsInDatabase.java
@@ -1,0 +1,57 @@
+package com.hartwig.hmftools.patientdb;
+
+import static com.hartwig.hmftools.common.utils.config.CommonConfig.SAMPLE;
+import static com.hartwig.hmftools.patientdb.dao.DatabaseAccess.addDatabaseCmdLineArgs;
+import static com.hartwig.hmftools.patientdb.dao.DatabaseAccess.databaseAccess;
+
+import java.sql.SQLException;
+
+import com.hartwig.hmftools.patientdb.dao.DatabaseAccess;
+
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+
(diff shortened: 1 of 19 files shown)


patient-db/src/main/java/com/hartwig/hmftools/patientdb/CheckSampleExistsInDatabase.java
```diff
@@ -0,0 +1,57 @@
...
+        try(DatabaseAccess dbAccess = databaseAccess(cmd))
+        {
+            if(dbAccess.readPurpleSampleList().contains(sample))
+            {
+                LOGGER.info("Sample '{}' exists in the database", sample);
+            }
+            else
+            {
+                LOGGER.info("Sample '{}' does not exist in the database", sample);
+                System.exit(2);
```
stefvanlieshout, line 39: Do you want to exit(2) in case a sample does not exist? It is a normal use-case for a sample not to exist (every first load).

stefvanlieshout (commented):
Nice improvement! One comment about an exit I'm not sure about though.

patient-db/src/main/java/com/hartwig/hmftools/patientdb/CheckSampleExistsInDatabase.java
mdenburger: I specifically added a special exit code to make this new class easier to use by automated tooling. Checking an exit code is a lot simpler than, for example, capturing and parsing the output.

The value `2` is a bit arbitrary, but aligns with the standard error code for "no such file or directory" ([ENONENT](https://www.baeldung.com/linux/standard-error-eodes)), which seemed a similar case. I wanted to keep `1` to indicate a real error, so `2` seemed logical as "one more option besides 1".

stefvanlieshout: Clear. Maybe "invert" the check though, such that an exit code is thrown in the case that a sample does exist, instead of when it does not. I'll approve anyway.

stefvanlieshout (approved):

mdenburger:
Good idea to revert the check, I've changed it to `CheckSampleAbsentFromDatabase`.

PR #526, opened 2024-02-23
DEV-3766: Fix germline MVLH table in PDF

DFKoetsier:
Changes:
- Remove unneeded division by 100 from when the input file contained percentages instead of factors.
- There are genes that have MVLH 100% and depth 0 in the input file due to being ignored by SAGE germline entirely. Remove these genes from both the MVLH table in the PDF and the related field in the JSON.

Please feel free to add more reviewers if you want.

I couldn't think of a better way to do the test I added to OrangeAlgoTest. If you can, please let me know.

### DIFF
--- a/orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
+++ b/orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
@@ -11,6 +11,7 @@ import java.text.DecimalFormat;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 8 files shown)


orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -368,38 +369,74 @@ private static EnsemblDataCache loadEnsemblDataCache(@NotNull OrangeConfig confi
...
+        Map<String, Double> mvlhPerGene = Maps.newTreeMap();
         for(String line : lines.subList(1, lines.size()))
         {
             String[] values = line.split(TSV_DELIM);
             String gene = values[geneIndex];
-            String mvlhString = values[mvlhIndex].substring(0, values[mvlhIndex].length() - 1);
-            double missedVariantLikelihood = Double.parseDouble(mvlhString) / 100D;
-            mvlhPerGene.put(gene, missedVariantLikelihood);
-        }
+            Double mvlh = Double.parseDouble(values[mvlhIndex]);
```
kduyvesteyn, line 405: Can be a `double` rather than `Double`

kduyvesteyn (approved):
Looks all good to me! Didn't test the actual MVLH table in the PDF though. 

Also, might be an idea to extract the MVLH logic out of OrangeAlgo and into a separate class. OrangeAlgo is mainly meant for tying together the various algo's, not really to execute anything non-trivial. Could also be for further improvement.

kzuberihmf (approved):
LGTM!

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
DFKoetsier: Agreed

DFKoetsier:
I have moved the MVLH logic from OrangeAlgo into a separate class

PR #527, opened 2024-02-23
DEV-3767: Update test run in Orange resource files to 5.34

DFKoetsier:
A simple resource update except for one small problem:
- I didn't have CUPPA output that includes RNA.
- The existing CUPPA loading tests relied on having RNA features present.
- The data already in the repo still had the bug where variants are not loaded correctly.

As a solution, I just put the new 5.34 CUPPA output without RNA in the "test_run" directory and the test reports. I turned the existing CUPPA file into a separate resource file for the existing tests, and I added tests for loading the CUPPA output file in "test_run" with only DNA features.


orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
```diff
@@ -12,22 +14,114 @@
...
     private static final double EPSILON = 1.0E-10;
 
-    private static final String CUPPA_VIS_DATA_TSV = Resources.getResource("test_run/cuppa/tumor_sample.cuppa.vis_data.tsv").getPath();
+    private static final String CUPPA_VIS_DATA_WITH_RNA_TSV = Resources.getResource("cuppa/with_rna.vis_data.tsv").getPath();
+    private static final String CUPPA_VIS_DATA_WITHOUT_RNA_TSV =
+            Resources.getResource("test_run/cuppa/tumor_sample.cuppa.vis_data.tsv").getPath();
 
     @Test
-    public void canExtractPredictionsFromCuppaV2() throws Exception
+    public void canExtractPredictionsFromCuppaV2IncludingRna() throws Exception
```
kduyvesteyn, line 30: Would remove the reference to CUPPA version in code. 

When we test a function related to purple we also don't use "purpleV3"!

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
kduyvesteyn, line 1: I'd be fine to leave the code as-is but strictly speaking: CuppaPredictions.fromTsv lives in another module, and shouldn't be invoked in this unit test. The CuppaDataFactory should be fed with in-memory testdata.

kduyvesteyn (approved):
Looks good! 

One question for double-check: Did lilac output not change between v5.33 and v5.34?

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
DFKoetsier: Agreed

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
DFKoetsier: The only real user of this class also used CuppaPredictions.fromTsv to load the prediction data.

I have slightly refactored CuppaDataFactory to include the call to CuppaPredictions.fromTsv. The only real effect is that the log line now counts the 40 tumor types instead of the number of lines of the cuppaVisDataTsvFile.

DFKoetsier:
> Looks good!
> 
> One question for double-check: Did lilac output not change between v5.33 and v5.34?

Thanks! And Lilac indeed did not change.

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
kduyvesteyn: Looks good to me

kzuberihmf (approved):

PR #528, opened 2024-02-27
ACTIN-642: Improve CUPPA output layout

cbruel:
* Remove underscores
* Rename GEN_POS to GENOMIC POSITION & GENE_EXP to GENE EXPRESSION
* Make clear that numbers behind drivers are driver likelihoods, and numbers behind "is male" and "whole genome duplication" are booleans

(Haven't made any changes to the SNV signatures highlighting yet, see mail)

### DIFF
--- a/cuppa/src/main/python/pycuppa/cuppa/visualization/plot_predictions.R
+++ b/cuppa/src/main/python/pycuppa/cuppa/visualization/plot_predictions.R
@@ -69,6 +69,12 @@ MAPPINGS_CLASSIFIER_NAMES = c(
   "GEN POS"="GENOMIC POSITION"
 )
 
+MAPPING_EVENT_NAMES = c(
+  "driver"="driver (DL)",
+  "virus"="virus (DL)",
+  "fusion"="fusion (DL)"
+)
+
 ## Heatmap of probs ================================
 get_plot_data_probs <- function(VIS_DATA, data_value_rounding = 2){
 
@@ -140,7 +146,7 @@ get_plot_data_probs <- function(VIS_DATA, data_value_rounding = 2){
       NA,
       "darkgrey"
    )
-   
+
    return(plot_data)
 }
 
@@ -344,14 +350,14 @@ plot_heatmap <- function(
 
 plot_probs <- function(VIS_DATA){
    plot_data <- get_plot_data_probs(VIS_DATA)
-   
+
    LOW_SNV_COUNT_THRES <- 50
    if(SNV_COUNT < LOW_SNV_COUNT_THRES){
       warning_string <- sprintf(" | WARNING: Sample has <%s SNVs. Predictions may be unreliable", LOW_SNV_COUNT_THRES)
    } else {
       warning_string <- ""
    }
-   
+
    plot_heatmap(
       plot_data,
       show_first_row_hline = TRUE,
@@ -486,7 +492,7 @@ plot_signatures <- function(
    scale_fill_manual(
       values=structure(quantile_info$colors[quantile_info$colors != "grey95"], names=quantile_info$labels[quantile_info$labels != "0"]),
       guide=guide_legend(override.aes=list(colour="black", linetype=1)),
-      na.value="grey95", drop=FALSE
+      na.value="grey95", drop=TRUE
    ) +
    labs(fill="Quantile in subtype cohort", title="SNV96: Mutational signatures")
 }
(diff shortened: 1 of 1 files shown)


cuppa/src/main/python/pycuppa/cuppa/visualization/plot_predictions.R
```diff
@@ -67,7 +67,9 @@ get_plot_data_probs <- function(VIS_DATA, data_value_rounding = 2){
 
    plot_data <- subset(VIS_DATA, data_type=="prob")
 
-   plot_data$row_label <- toupper(plot_data$clf_name)
+   plot_data$row_label <- toupper(gsub("_"," ", plot_data$clf_name))
+   plot_data$row_label[plot_data$row_label == "GENE EXP"] <- "GENE EXPRESSION"
+   plot_data$row_label[plot_data$row_label == "GEN POS"] <- "GENOMIC POSITION"
```
luan-n-nguyen, line 72: It would be better to define mappings that the top of the script, e.g.:

```
MAPPINGS_CLASSIFIER_NAMES = c(
  GENE_EXP="GENE EXPRESSION",
  GEN_POS="GENOMIC POSITION"
)
```

and then use a function like `revalue` from `plyr` (or write a custom function):
`plyr::revalue(plot_data$row_label, replace=MAPPINGS_CLASSIFIER_NAMES)`
See: https://www.rdocumentation.org/packages/plyr/versions/1.8.9/topics/revalue

I would also define `MAPPINGS_FEAT_NAMES`

cbruel: @luan-n-nguyen Thanks for the tip on the revalue function! 

Re `MAPPINGS_FEAT_NAMES`: Do you mean to explicitly map every feature name to its non underscored version instead of using `gsub("_", " ", ..)`?
```
MAPPING_FEAT_NAMES = c(
      snv_count="snv count"
      is_male="is male",
      whole_genome_duplication="whole genome duplication",
      ....
)
```

luan-n-nguyen: Oh never mind this comment. For the feature names it's indeed better to use some substitutions / regex rules

cuppa/src/main/python/pycuppa/cuppa/visualization/plot_predictions.R
```diff
@@ -515,7 +524,7 @@ plot_feat_contrib <- function(VIS_DATA){
       feat_value_label[feat_value==0] <- "0"
       feat_value_label[feat_value==1] <- "1"
 
-      paste0(affixes$feat_basename, " = ", feat_value_label)
+      paste0(ifelse(grepl("\\.", affixes$feat_basename), paste(gsub("\\.", " ", affixes$feat_basename), " (", feat_value_label, ")", sep = ""), ifelse((affixes$feat_basename == "is male") | (affixes$feat_basename == "whole genome duplication"), paste(affixes$feat_basename, "=", feat_value_label, sep = " "), paste(affixes$feat_basename, "=", feat_value_label, sep = " "))))
```
luan-n-nguyen, line 527: This could be split into several steps to make it easier to read.

Also `(affixes$feat_basename == "is male") | (affixes$feat_basename == "whole genome duplication")` can be replaced with `affixes$feat_basename %in% c("is male", "whole genome duplication")`

cbruel:
@luan-n-nguyen One more thing we would like to clarify in the visualization are the driverlikehoods 0 e.g. "APC mut (0)". This could indicate either an APC variant with driverlikelihood 0 ór the absence of an APC variant? In the latter we could change it to "APC mut (absent)" to emphasize the difference. 

Is it possible to distinguish the 2 cases based on the `feat_value`? I can find that absent variants have a `feat_value` of 0.0, but do you know what would be the `feat_value` in case a variant has a driverlikehood of 0%, is this different from 0.0?

luan-n-nguyen:
As of right now there is no way to tell the difference.
There would need to be an update to CUPPA to make the distinction (e.g. encode absence with -0.001).

cbruel:
@luan-n-nguyen Thanks for confirmation! Then we will leave that for future improvement

cbruel:
@luan-n-nguyen Absent drivers (feat_value = -1e-8) are now shown with `"(absent)"`, instead of `"(0)"`. I didn't think it was necessary to change anything to the number rounding logic in the script, but please let me know if you think that is still needed

PR #534, opened 2024-03-07
DEV-3771: Bam-tools metrics ignores some reads for WgsMetrics

DFKoetsier:
- Consensus reads are ignored for WgsMetrics. 
- When running in target-only mode, reads starting just before a region are never counted as coverage for that region.

I also noticed some slight inconsistencies in the treatment of secondary alignments that I have tried to fix.


DFKoetsier:
Already fixed in different branch and merged into master.

PR #539, opened 2024-04-02
ACTIN-551: Add maxCopyNumber to PurpleGeneCopyNumber datamodel in orange datamodel

dvanbeek:


DFKoetsier (changes requested):
The changes so far look good.

I would prefer to include the changes in Orange in the same PR, because otherwise hmftools won't successfully build because of failing Orange tests.

Also, I think it would be good to uptick the Orange datamodel version in the pom, since you are making a change to it. Especially because the change is not backwards compatible.

PR #543, opened 2024-04-17
ACTIN-808: Add PurpleTranscriptImpact.reported flag to orange datamodel

kzuberihmf:
* PurpleTranscriptImpact gets reported boolean (can not be null)
* PupleVariantFactory uses PurpleVariantContext.reportrableTranscripts() to set the reported flag for other impacts
* PurpleVariant.reported is derived, and true if canonical or any other transcripts are reported

Hope this matches the intent discussed. 

Let's suppose the next version is 3.5.0, with this change it won't be able to load orange-datamodel jsons from 3.4.0 since the reported field is required. Are we ok with this break in compatibility?


orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/PurpleVariant.java
```diff
@@ -68,4 +66,18 @@ public interface PurpleVariant
...
+    {
+        if(canonicalImpact().reported())
+        {
+            return true;
+        }
+        else
+        {
+            return otherImpacts().stream().anyMatch(PurpleTranscriptImpact::reported);
+        }
+    }
```
DFKoetsier, line 82: Do we want this to be deprecated? I thought the idea was to keep this longer term, since otherwise we would have to reimplement the method in multiple data-model users.

DFKoetsier (changes requested):
Looks good! I had only a few remarks.

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/PurpleVariant.java
DFKoetsier: Also, Orange uses this field itself. If we do want to deprecate it, shouldn't we already remove those calls in this pull request?

kduyvesteyn: I'd lean towards deprecation, but no strong opinion. So far the ORANGE datamodel has been as lean-and-mean as possible, meaning there is no room for derived values inside the datamodel. 

And agree on the 2nd item!

orange-datamodel/src/test/resources/real.orange.json
```diff
@@ -163,7 +164,8 @@
           "effects": [
             "MISSENSE"
           ],
-          "codingEffect": "MISSENSE"
+          "codingEffect": "MISSENSE",
+          "reported": true
         },
```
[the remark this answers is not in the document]
kduyvesteyn: This file was created from COLO using `ReportLimiter`. This logic doesn't guarantee a self-consistent dataset, but think that is not necessary either. That being said, the name could be improved, or we could even  decide to merge with minimally populated or delete it, it doesn't really add a lot. At this point, the minimally populated is manually managed, while the real is generated from actual data. 

I don't have a strong opinion on what should change (if anything), but would be good to generate the `real.orange.json` from the actual Orange Test Application.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -119,4 +123,85 @@ public void testCanCreateVariantFromContext()
         assertEquals(1, purpleVariant.subclonalLikelihood(), 0);
         assertEquals(List.of(1, 2, 3), purpleVariant.localPhaseSets());
     }
+
+    @Test
+    public void testDetermineReportableTranscripts()
```
kduyvesteyn, line 128: I feel you can slim down this test (and use `TestPurpleVariantFactory` builders). This test currently tests way more than just the reportable transcripts.

kduyvesteyn (commented):
Looks good! One minor comment in addition to David's earlier findings.

kduyvesteyn:
> * PurpleTranscriptImpact gets reported boolean (can not be null)
> * PupleVariantFactory uses PurpleVariantContext.reportrableTranscripts() to set the reported flag for other impacts
> * PurpleVariant.reported is derived, and true if canonical or any other transcripts are reported
> 
> Hope this matches the intent discussed.
> 
> Let's suppose the next version is 3.5.0, with this change it won't be able to load orange-datamodel jsons from 3.4.0 since the reported field is required. Are we ok with this break in compatibility?

My understanding is that a new field doesn't break compatibility? The only thing we need to keep in mind is that we need to upgrade ORANGE in the pipeline before we bump the datamodel dependency in ACTIN (to make sure we produce the datamodel that ACTIN expects). Please double-check with Paul to be sure!

orange-datamodel/src/test/resources/real.orange.json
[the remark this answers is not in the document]
kzuberihmf: Oh, I see. I just ran this test application, presumably all the (unrelated to this PR) differences I see in the output are to be expected, e.g. platinum version 5.34 etc, I think all from the test files that must have been updated since the previous time real.orange.json was generated.

If that sounds right, I can check this in (after running jq to for layout). One odd thing is that the current version has unicode characters in the variants for the > char, wonder how that happened 🤔. The version I generated seems ok though.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactory.java
```diff
@@ -89,6 +91,7 @@ private PurpleTranscriptImpact extractCanonicalImpact(PurpleVariantContext purpl
                 .inSpliceRegion(purpleContext.spliceRegion())
                 .effects(purpleVariantEffects)
                 .codingEffect(PurpleConversion.convert(purpleContext.canonicalCodingEffect()))
+                .reported(purpleContext.reported())
```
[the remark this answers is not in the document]
kzuberihmf: Good catch! Thx.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
kzuberihmf: Simplified the tests, thanks.

kzuberihmf:
> 
> My understanding is that a new field doesn't break compatibility? The only thing we need to keep in mind is that we need to upgrade ORANGE in the pipeline before we bump the datamodel dependency in ACTIN (to make sure we produce the datamodel that ACTIN expects). Please double-check with Paul to be sure!

Tested this by creating an Orange json using this version on COLO and processing it through a recent version of ACTIN-Molecular and it ran without error. So seems for that direction (will also mention to Paul).

orange-datamodel/src/test/resources/real.orange.json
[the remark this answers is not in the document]
kduyvesteyn: Yes, would be good to refresh this file with all changes included.

orange/README.md
```diff
@@ -231,14 +231,17 @@ investigate potential causes for QC failure.
 
 ## Version History and Download Links
 
+- [upcoming]
```
kduyvesteyn, line 234: Minor but wouldn't enclose in brackets (there is no link anyways)

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantContextLoader.java
```diff
@@ -179,6 +186,15 @@ private static AllelicDepth extractRnaDepth(VariantContext context, @Nullable St
                 .orElse(null);
     }
 
+    @NotNull
+    private static List<String> extractReportableTranscripts(VariantContext context)
```
kduyvesteyn, line 190: Another minor but we typically use nullable annotations for every function input and output in ORANGE, so `context` should be annotated as `@NotNull`

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -119,4 +124,83 @@ public void testCanCreateVariantFromContext()
         assertEquals(1, purpleVariant.subclonalLikelihood(), 0);
         assertEquals(List.of(1, 2, 3), purpleVariant.localPhaseSets());
     }
+
+    @Test
+    public void testReportableTranscripts()
+    {
+        ImmutablePurpleVariantContext context = TestPurpleVariantFactory.contextBuilder()
```
kduyvesteyn, line 131: The type can just be `PurpleVariantContext`

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -119,4 +124,83 @@ public void testCanCreateVariantFromContext()
...
+
+        PurpleTranscriptImpact secondImpact = purpleOtherImpacts.get(1);
+        assertEquals("other_transcript2", secondImpact.transcript());
+        assertFalse(secondImpact.reported());
+    }
+
+    @Test
+    public void testCanonicalTranscriptNotReported()
+    {
+        ImmutablePurpleVariantContext context = TestPurpleVariantFactory.contextBuilder()
```
kduyvesteyn, line 164: Type can just be `PurpleVariantContext`

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -119,4 +124,83 @@ public void testCanCreateVariantFromContext()
...
+                .canonicalTranscript("canonical_transcript")
+                .reported(false)
+                .otherImpacts(List.of(transcriptImpact("other_transcript1"), transcriptImpact("other_transcript2")))
+                .reportableTranscripts(Set.of("other_transcript1"))
+                .build();
+
+        PaveAlgo paveAlgo = new PaveAlgo(TestEnsemblDataCacheFactory.createDummyCache(), false);
+        PurpleVariant purpleVariant = new PurpleVariantFactory(paveAlgo).fromPurpleVariantContext(context);
```
kduyvesteyn, line 172: Not sure but might be nice as test here to assert that purpleVariant.reported is true?

Might be that this is tested in another place, in that case ignore!

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
```diff
@@ -119,4 +124,83 @@ public void testCanCreateVariantFromContext()
...
+        PurpleTranscriptImpact firstImpact = purpleOtherImpacts.get(0);
+        assertEquals("other_transcript1", firstImpact.transcript());
+        assertTrue(firstImpact.reported());
+
+        PurpleTranscriptImpact secondImpact = purpleOtherImpacts.get(1);
+        assertEquals("other_transcript2", secondImpact.transcript());
+        assertFalse(secondImpact.reported());
+    }
+
+    private VariantTranscriptImpact transcriptImpact(String transcript)
```
kduyvesteyn, line 194: Should be annotated with `@NotNull` and should be declared `static`. No kotlin code!!

kduyvesteyn (commented):
One additional thought: I suspect `VariantEntryFactory` in ORANGE can be improved thanks to the changes in the datamodel. This function constructs the actual variants as they are expected to appear on the report, and this is derived from drivers since variant objects themselves did not contain enough information.

With this change I suspect the VariantEntry list that is generated in the `VariantEntryFactory` can be derived solely from purple variants (potentially still need drivers for additional annotation but not as the source for variant entries... didn't think through completely though.

DFKoetsier:
> One additional thought: I suspect `VariantEntryFactory` in ORANGE can be improved thanks to the changes in the datamodel. This function constructs the actual variants as they are expected to appear on the report, and this is derived from drivers since variant objects themselves did not contain enough information.
> 
> With this change I suspect the VariantEntry list that is generated in the `VariantEntryFactory` can be derived solely from purple variants (potentially still need drivers for additional annotation but not as the source for variant entries... didn't think through completely though.

I'm pretty sure you'd still need the drivers for the driver likelihood. I do agree that it's probably possible to use them for only that.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/PurpleVariantFactoryTest.java
kzuberihmf: Yeah I totally got out of the habit of thinking of annotations after being in kotlin code for a while!

DFKoetsier (approved):
My comments have been addressed. Looks great!

kzuberihmf:
Will merge this, and do VariantEntryFactory in a separate PR. 

Shall I do a release for ORANGE and ORANGE datamodel for this now as well?

PR #552, opened 2024-05-06
PatientDB: restore index prefix length of cpctEcrf.itemValue column

mdenburger:
MySQL needs an index prefix limit for the `cpctEcrf.itemValue` column, otherwise creating the index fails. Since jOOQ cannot parse the index prefix limit we ignore this index during jOOQ code generation. The index prefix limit value 255 [was used before](https://github.com/hartwigmedical/hmftools/commit/38ba9b11ef59d8ebcccda3ed8956a9d7e92e0f4f#diff-6c4cf5de30b233ee5dc200058d99b17591b120b9793a2ae77a3098611e599d82L219).


KasperWolsink99 (approved):
nice! I am not sure if this is possible but maybe we can ignore indexes in general by changing the codegen configuration in the pom?

mdenburger:
> nice! I am not sure if this is possible but maybe we can ignore indexes in general by changing the codegen configuration in the pom?

I could not find a way to ignore indexes in general. The ignore comments seemed the simplest fix with the least amount of changes.

mdenburger:
The clinical data loading is currently being moved out of hmftools, so this change will be obsolete quickly. Deleting this PR.

PR #560, opened 2024-06-05
ACTIN-995: Add immune escape analysis to ORANGE datamodel and report

kduyvesteyn:
See JIRA for design and rationale.

Some notes on implementation:
 - I have checked that all genes that are configured exist in our genome (HGNC)
 - The LOH check in immune escape is inconsistent (MACN < 0.3) with other LOH algo's in ORANGE (MACN < 0.5). The current GIE implementation is consistent with the GIE paper and vignette though, and according to Fran deliberately chosen to be conservative.
 - I have chosen for the immune escape to be omni-present (NotNull). I feel this is the fine choice, but does require some considerations:
    - If some genes that are checked in immune escape analysis but were not part of the panel that went into ORANGE, we will have missing data. Have decided to go with a WARN in this case, but may not be the cleanest. 
    - If QC is e.g. `FAIL_NO_TUMOR`, it may be cleaner to have null immune escape instead of false everywhere. However, "false" is still not a wrong value in this case, so want to leave up to downstream (and in theory we could still find something in case of non-PASS QC and don't want to filter). 
        - Note that in the report I blank out the immune escape table in case of QC failure!

Have added a screenshot of the new immunology page of COLO829 report for ease of review of the reporting changes.

immunology_COLO829


orange/src/main/java/com/hartwig/hmftools/orange/algo/immuno/ImmuneEscapeInterpreter.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private static boolean hasLOH(@NotNull List<PurpleGeneCopyNumber> allSomaticGeneCopyNumbers, @NotNull String geneToCheck)
+    {
+        for(PurpleGeneCopyNumber somaticGeneCopyNumber : allSomaticGeneCopyNumbers)
+        {
+            if(somaticGeneCopyNumber.gene().equals(geneToCheck))
+            {
+                double minMajorAlleleCopyNumber = somaticGeneCopyNumber.minCopyNumber() - somaticGeneCopyNumber.minMinorAlleleCopyNumber();
+                return somaticGeneCopyNumber.minMinorAlleleCopyNumber() < 0.3 && minMajorAlleleCopyNumber > 0.7;
```
DFKoetsier, line 66: Does the name minMajorAlleleCopyNumber actually make sense? Since the minCopyNumber and minMinorAlleleCopyNumber can be from different copy number regions, "minMajorAlleleCopyNumber" doesn't have to be the major allele copy number anywhere. For instance if the gene partially has CN==3 with minorAlleleCN==0 and partially CN==2 and minorAlleleCN==1, then this calculation gives 2, which isn't the major allele copy number anywhere.

Also, if we only call deletions when minCN<0.5, and can only call LOH when minCN>0.7+minMinorAlleleCN. Then if minMinorAlleleCN is 0 and minCN is 0.6, we call neither LOH nor a deletion, which is unfortunate since the only actual unclear bit in that situation is which of the two it is. And if minMinorAlleleCN is greater than 0, then the minCN range where we can call neither deletion nor LOH increases. I think it isn't great that we would miss the inactivation in that case.

I think we should consider replacing the minMajorAlleleCopyNumber part of the check with `somaticGeneCopyNumber.minCopyNumber() >= 0.5`. 

The only counterargument to this I can think of is that perhaps in that in between region we think it's a subclonal deletion, and that we don't want to call that as escape since it isn't clonal. But it doesn't sound like a great argument to me, and it definitely doesn't apply when minMinorAlleleCN is exactly 0.

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeRecord.java
```diff
@@ -69,6 +70,9 @@ public interface OrangeRecord
     @NotNull
     LilacRecord lilac();
 
+    @NotNull
+    ImmuneEscapeRecord immuneEscape();
+
```
DFKoetsier, line 75: We might want to make this field Nullable so we can handle tumor-only runs cleanly. If we don't null this for tumor-only runs then we could end up calling too many inactivations due to biallelic germline missense mutations that are harmless.

Or we'd have to do something smarter in tumor-only runs.

DFKoetsier (commented):
Looks pretty good! Just a couple of questions and remarks.

orange/src/test/java/com/hartwig/hmftools/orange/report/ReportGeneratorTestApplication.java
```diff
@@ -45,8 +46,10 @@ public class ReportGeneratorTestApplication
 
     private static final boolean USE_MOCK_DATA_FOR_REPORT = false;
     private static final boolean REMOVE_UNREPORTED_VARIANTS = false;
+
+    // Switch LIMIT_JSON_OUTPUT to true if you want to generate the real.orange.json test resource in orange-datamodel!
     private static final boolean LIMIT_JSON_OUTPUT = false;
-    private static final Set<PurpleQCStatus> OVERRIDE_QC_STATUS = null;
+    private static final Set<PurpleQCStatus> OVERRIDE_QC_STATUS = Sets.newHashSet(PurpleQCStatus.FAIL_NO_TUMOR);
```
kzuberihmf, line 52: Did you mean to commit this change?

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/ImmuneEscapeTable.java
```diff
@@ -0,0 +1,61 @@
...
+public final class ImmuneEscapeTable
+{
+    @NotNull
+    public static Table build(@NotNull String title, float width, @NotNull ImmuneEscapeRecord immuneEscape,
+            @NotNull ReportResources reportResources, boolean isTumorFail)
+    {
+        Cells cells = new Cells(reportResources);
+        Table table = Tables.createContent(width,
+                new float[] { 2, 1, 3 },
+                new Cell[] { cells.createHeader("Escape Mechanism"), cells.createHeader("Present?"), cells.createHeader(Strings.EMPTY) });
```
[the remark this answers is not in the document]
kduyvesteyn: I see your point, agree "present" is too definitive. I have changed it to "detected?" which I think covers the analysis better. If something is not detected then it could still be present in theory. 

I do think it is nice to display all 6 categories to make it transparant what we actually look for in every report. 

Have also considered updating the datamodel based on your feedback but I think those names are fine, fairly consistent with other characteristics of a tumor we try to capture but with <100% sensitivity.

orange-datamodel/src/test/resources/real.orange.json
```diff
@@ -154,7 +154,7 @@
         "worstCodingEffect": "MISSENSE",
         "canonicalImpact": {
           "transcript": "ENST00000335508",
-          "hgvsCodingImpact": "c.2153C>T",
+          "hgvsCodingImpact": "c.2153C\u003eT",
```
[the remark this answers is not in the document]
kduyvesteyn: I have regenerated the `real.orange.json` by running ReportGeneratorTestApplication with `LIMIT_JSON_OUTPUT` set to true. Not sure why I generate these unicode's locally, but have corrected them manually now, thanks for spotting!

orange-datamodel/src/test/resources/real.orange.json
```diff
@@ -23,8 +23,8 @@
       "pctExcAdapter": 0.0,
       "pctExcMapQ": 0.04985,
       "pctExcDupe": 0.10868,
-      "pctExcUnpaired": 0.00071,
-      "pctExcBaseQ": 0.000040,
+      "pctExcUnpaired": 7.1E-4,
+      "pctExcBaseQ": 4.0E-5,
```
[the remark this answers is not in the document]
kduyvesteyn: See other comment, not entirely sure why this happens, though I imagine it must be some local setting on my laptop that is different from the previous update. I'll leave this as-is since it seems fine.

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/ImmuneEscapeTable.java
```diff
@@ -0,0 +1,61 @@
...
+
+        table.addCell(cells.createContent("(Potential) PD-L1 overexpression"));
+        table.addCell(cells.createContent(toYesNoUnavailable(immuneEscape.hasPDL1OverexpressionEscape(), isTumorFail)));
+        table.addCell(cells.createContent(Strings.EMPTY));
+
+        table.addCell(cells.createContent("CD58 inactivation"));
+        table.addCell(cells.createContent(toYesNoUnavailable(immuneEscape.hasCD58InactivationEscape(), isTumorFail)));
+        table.addCell(cells.createContent(Strings.EMPTY));
+
+        table.addCell(cells.createContent("Epigenetic driven immune escape via SETDB1"));
```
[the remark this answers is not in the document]
kduyvesteyn: Have rephrased to "Epigenetics driven"

orange/src/main/java/com/hartwig/hmftools/orange/algo/immuno/ImmuneEscapeInterpreter.java
kduyvesteyn: Good suggestion. I've tried to stay as close to the vignette implementation as possible, but didn't feel exactly right. This is better, and will let Joseph know I've changed this. 

Did make it ">", I am not a huge fan of using = for comparing doubles, check for LOH now is:

`somaticGeneCopyNumber.minCopyNumber() > 0.5  && somaticGeneCopyNumber.minMinorAlleleCopyNumber() < 0.3`

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangeRecord.java
kduyvesteyn: Would prefer to leave as-is right now, as I think we should be able to say something sensible about immune escape in tumor-only. The current implementation is shaky in the end but maybe the problem is upstream (poor filtering of germline variants) or with the GIE implementation in general. Will let Joseph know

kduyvesteyn:
Note: Haven't regenerated the example reports for the changes in the report itself, will do so after merging to master.

DFKoetsier (approved):
Looks good!

kzuberihmf (approved):

PR #561, opened 2024-06-05
ACTIN-999: Remove pre-LILAC v1.6 check from ORANGE

kduyvesteyn:
LILAC pre v1.6 used to write fragment counts to ref fields in case it was run in tumor-only mode. This has been fixed in LILAC v1.6, and since we depend on that now, we can remove the work-around for this in ORANGE.

With this change, ORANGE assumes tumor fragments are filled in when run in tumor-only mode. 

Have tested using COLO829 tumor+ref and tumor-only.


DFKoetsier (approved):
Looks good!

kzuberihmf (approved):

PR #562, opened 2024-06-09
ACTIN-1007: Set context to "Unknown" for promiscuous enhancer targets

kduyvesteyn:
Note: Have changed the catch-all in the context function to an exception rather than "ERROR" string since all current transcript region types are now covered.

For extra safety, added a unit test that will fail in case someone would add a new type of TranscriptRegionType.


kzuberihmf (approved):
LGTM

PR #567, opened 2024-06-14
Switch to MIT license

kduyvesteyn:
Would appreciate if you could review the license and be sure about this change!


kduyvesteyn:
Closed because MIT would probably violate some upstream licenses used by hmftools

PR #569, opened 2024-06-21
ACTIN-1057: Filter NaN likelihoods from CUPPA when converting in ORANGE

kduyvesteyn:
Have tested on data-vm and confirmed this ORANGE version creates a JSON file which can be ingested into ACTIN (no NaN exception anymore).

### DIFF
--- a/orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactory.java
+++ b/orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactory.java
@@ -29,7 +29,7 @@ public final class CuppaDataFactory
     @NotNull
     private static CuppaData createFromPredictions(@NotNull CuppaPredictions cuppaPredictions) throws Exception
     {
-        List<CuppaPrediction> predictions = extractSortedProbabilities(cuppaPredictions);
+        List<CuppaPrediction> predictions = convertCuppaPredictions(cuppaPredictions);
 
         if(predictions.isEmpty())
         {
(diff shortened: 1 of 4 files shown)


orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactory.java
```diff
@@ -83,9 +79,13 @@ private static List<CuppaPrediction> extractSortedProbabilities(@NotNull CuppaPr
                     .altSjCohortClassifier(probabilitiesByClassifier.get(ClassifierName.ALT_SJ))
                     .build();
 
-            cuppaPredictionsOrangeFormat.add(prediction);
+            // If a classifier has no data for a specific cancer type we should remove it completely, see ACTIN-1057
```
kzuberihmf, line 82: Feels a bit unusual to refer to a ticket in code, unless its an open TODO that needs followup.

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
```diff
@@ -58,11 +58,12 @@ public void canExtractPredictionsFromCuppaIncludingRna() throws Exception
                 .build();
 
         Map<String, CuppaPrediction> expectedPredictionsByCancerType = new HashMap<>();
-        expectedPredictionsByCancerType.put("Skin: Melanoma", expectedPredictionMelanoma);
-        expectedPredictionsByCancerType.put("Skin: Other", expectedPredictionSkinOther);
-        expectedPredictionsByCancerType.put("Prostate", expectedPredictionProstate);
+        expectedPredictionsByCancerType.put(expectedPredictionMelanoma.cancerType(), expectedPredictionMelanoma);
+        expectedPredictionsByCancerType.put(expectedPredictionSkinOther.cancerType(), expectedPredictionSkinOther);
+        expectedPredictionsByCancerType.put(expectedPredictionProstate.cancerType(), expectedPredictionProstate);
+        expectedPredictionsByCancerType.put("Bone/Soft tissue: Cartilaginous neoplasm", null);
```
kzuberihmf, line 64: Where is the null setup in the test data? ... oh ok, I see that it was always there but its actually missing values in the input file from CUPPA and not explicit NaN's.

kzuberihmf (approved):

orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactory.java
pauldwolfe: indeed I'd normally just include this sort of thing in the commit message. Future devs can annotate the code to see it.

pauldwolfe (approved):

orange/src/test/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactoryTest.java
kduyvesteyn: Correct, that's what tricked us in the first place :)

orange/src/main/java/com/hartwig/hmftools/orange/algo/cuppa/CuppaDataFactory.java
kduyvesteyn: Updated

kduyvesteyn: Do note that I added a comment because the implementation choice may not be optimal. We could for instance also expose a wider CUPPA datamodel in ORANGE and leave to downstream how to handle NaN likelihoods, but considered that unnecessary at this point

PR #580, opened 2024-07-10
INFRA-408 add back snpcheck and remove lucene dependency

mvanniekerkHartwig:
Adding back snpcheck should fix the build. Also removing lucene as suggested by Korneel, since it is now unused in the project.


kduyvesteyn (approved):
Looks good! 

Confirmed local build works with these changes.

stefvanlieshout (approved):
Like mentioned earlier, I think the snpcheck table should nog be required anymore, but keeping it for now not an issue of course.

PR #582, opened 2024-07-11
INFRA-585 create a dockerfile for hmf id generator

mvanniekerkHartwig:


hmf-id-generator/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM eclipse-temurin:11-jre
+COPY target/hmf-id-generator-3.1-jar-with-dependencies.jar /usr/share/hmf-id-generator.jar
+ENTRYPOINT ["java", "-jar", "/usr/share/hmf-id-generator.jar"]
```
mdenburger, line 3: Should we create a regular user in Dockerfiles too and run as that user? That avoids the Kubernetes warning `Warning: would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false`.

The recipe is quite simple:

Create a user with 
```
RUN useradd --create-home --shell /bin/bash --no-log-init user
WORKDIR /home/user
```
and run as that user with (before ENTRYPOINT):
```
USER user
```

In a deployment, use
```
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["ALL"]
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
```

mdenburger (approved):

hmf-id-generator/Dockerfile
mdenburger: Ah I see we only run this from HAWE, so I guess a user is not very necessary here. Still good to know (and good to apply to new Dockerfiles that will run in k8s I guess).

PR #583, opened 2024-07-12
ACTIN-96: Sort signatures by allocation and add etiology

cbruel:


hmf-common/src/test/java/com/hartwig/hmftools/common/sigs/SignatureTestFactory.java
kduyvesteyn, line 1: I would not adjust this test factory, this relates to the hmf-common datamodel (sigs) and not the orange datamodel.

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/sigs/SignatureAllocation.java
```diff
@@ -16,4 +16,7 @@ public interface SignatureAllocation
     double allocation();
 
     double percent();
+
+    @Nullable
```
kduyvesteyn, line 20: Why would this ever be null? 

I can imagine there is no etiology for `MISALLOC` but maybe you can assign "-" in the datamodel in that case? Since etiology is not for display purpose anyways I think I would avoid null

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -73,6 +73,7 @@ public interface OrangeConfig
     String DOID_JSON = "doid_json";
     String COHORT_MAPPING_TSV = "cohort_mapping_tsv";
     String COHORT_PERCENTILES_TSV = "cohort_percentiles_tsv";
+    String SIGNATURES_ETIOLOGIES_TSV = "signatures_etiologies_tsv";
```
kduyvesteyn, line 76: In general I believe the correct term is "signature_etiologies". 

There is 1 signature_etiology
There are N signature_etiologies 

:)

(affects naming in multiple places)

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -106,6 +107,8 @@ public class OrangeAlgo
     @NotNull
     private final List<DriverGene> driverGenes;
     @NotNull
+    private final Map<String, String> signaturesEtiologies;
```
kduyvesteyn, line 110: Maybe a more correct name would be "etiologyPerSignature". That makes it more clear your keys are signatures and the values are the etiologies?

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -150,18 +157,20 @@ public static OrangeAlgo fromConfig(@NotNull OrangeConfig config) throws IOExcep
         String outputDir = config.outputDir();
         PlotManager plotManager = !outputDir.isEmpty() ? new FileBasedPlotManager(outputDir) : new DummyPlotManager();
 
-        return new OrangeAlgo(doidEntry, mapper, percentilesModel, driverGenes, knownFusionCache, ensemblDataCache, plotManager);
+        return new OrangeAlgo(doidEntry, mapper, percentilesModel, driverGenes, signaturesEtiologies, knownFusionCache, ensemblDataCache, plotManager);
```
kduyvesteyn, line 160: Minor but this line extends max length, should be auto-reformatted.

orange/src/main/java/com/hartwig/hmftools/orange/algo/sigs/SigsEtiologiesLoader.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.Maps;
+
+import org.jetbrains.annotations.NotNull;
+
+public class SigsEtiologiesLoader
```
kduyvesteyn, line 16: Class should be final (not meant to be extended, its the equivalent of "object" in kotlin)

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/SignatureAllocationTable.java
```diff
@@ -72,10 +75,7 @@ else if(sig2.equalsIgnoreCase(MISALLOC_SIGNATURE))
                 return sig1.compareTo(sig2);
             }
 
-            int sigNumber1 = Integer.parseInt(sig1.substring(3));
-            int sigNumber2 = Integer.parseInt(sig2.substring(3));
-
-            return Integer.compare(sigNumber1, sigNumber2);
+            return Double.compare(allocation1.allocation(), allocation2.allocation()) * -1;
```
kduyvesteyn, line 78: It is more typical to use `Double.compare(allocation2, allocation1)` if you want to order descending.

orange/src/main/java/com/hartwig/hmftools/orange/conversion/OrangeConversion.java
```diff
@@ -154,13 +159,22 @@ public static PeachGenotype convert(@NotNull com.hartwig.hmftools.common.peach.P
                 .build();
     }
 
-    @NotNull
-    public static SignatureAllocation convert(@NotNull com.hartwig.hmftools.common.sigs.SignatureAllocation signatureAllocation)
+    @Nullable
+    public static List<SignatureAllocation> convertAndAnnotateWithEtiology(
+            @Nullable List<com.hartwig.hmftools.common.sigs.SignatureAllocation> signatureAllocation,
```
kduyvesteyn, line 164: Why can signatureAllocation be null?

orange/src/main/java/com/hartwig/hmftools/orange/conversion/OrangeConversion.java
kduyvesteyn, line 1: I would not put the the etiology annotation in this class. This class is meant for trivial conversion. 

I would suggest to make a new class "SigsInterpreter" that implements this logic

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/SignatureAllocationTable.java
```diff
@@ -33,13 +34,15 @@ public static Table build(@NotNull String title, float width, @NotNull List<Sign
...
-                new Cell[] { cells.createHeader("Signature"), cells.createHeader("Allocation"), cells.createHeader("Percent"),
+                new float[] { 1, 3, 1, 1, 1 },
+                new Cell[] { cells.createHeader("Signature"), cells.createHeader("Etiology"), cells.createHeader("Allocation"),
+                        cells.createHeader("Percent"),
                         cells.createHeader(Strings.EMPTY) });
 
         for(SignatureAllocation signatureAllocation : sort(signatureAllocations))
         {
             table.addCell(cells.createContent(signatureAllocation.signature()));
+            table.addCell(cells.createContent(Objects.requireNonNullElse(signatureAllocation.etiology(), "-")));
```
kduyvesteyn, line 45: As mentioned before, I would set "-" for MISALLOC in the algo and make the etiology itself NotNull

orange/src/test/java/com/hartwig/hmftools/orange/report/tables/SignatureAllocationTableTest.java
```diff
@@ -16,24 +16,24 @@ public class SignatureAllocationTableTest
...
-        assertEquals(allocation2, sorted.get(1));
+        assertEquals(allocation2, sorted.get(0));
+        assertEquals(allocation1, sorted.get(1));
         assertEquals(allocation3, sorted.get(2));
         assertEquals(allocation4, sorted.get(3));
     }
 
     @NotNull
-    private static SignatureAllocation create(@NotNull String signature)
+    private static SignatureAllocation create(@NotNull String signature, @NotNull Double allocation)
```
kduyvesteyn, line 35: `double` is the same as `NotNull Double`

orange/src/test/resources/sigs/signatures_etiologies.tsv
kduyvesteyn, line 1: I would put the entire file in the test resources.

Imagine in a future pipeline run COLO gets a signature that is not in this file - we would have to update it again.

Also, I would put this file in `test_run_resources` to make it clear it is there to support the `test_run` ORANGE report

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/sigs/SignatureAllocation.java
cbruel: Indeed my reasoning behind making it nullable was because of `MISALLOC`

orange/src/main/java/com/hartwig/hmftools/orange/conversion/OrangeConversion.java
cbruel: This input comes from `loadSigAllocations`, which could have a nullable output (this could e.g. happen when no sigs allocation tsv is provided)

kduyvesteyn: Clear! Didn't realise you changed the input from a single sig to a list of sigs, my bad!

hmf-common/src/test/java/com/hartwig/hmftools/common/sigs/SignatureTestFactory.java
cbruel: Reverted the change. But `SignatureTestFactory` is now not used anymore (now it isn't used anymore in `OrangeConversionTest`). So I would say the factory could be removed?

orange/src/test/resources/sigs/signatures_etiologies.tsv
cbruel: I've put the entire file in the test resources, and uploaded the file in `common-resources-public`.

orange/src/main/java/com/hartwig/hmftools/orange/algo/sigs/SigsEtiologiesLoader.java
cbruel: I've also added a `SigsEtiologiesLoaderTest` btw, since I thought it was missing

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/sigs/SignatureAllocation.java
```diff
@@ -16,4 +16,6 @@ public interface SignatureAllocation
     double allocation();
 
     double percent();
+
+    String etiology();
```
kduyvesteyn, line 20: Should be NotNull :)

Also, I think I would put this property under "signature" (same as what you have done in the example jsons, feels like natural order.

orange/src/main/java/com/hartwig/hmftools/orange/algo/sigs/SigsInterpreter.java
```diff
@@ -0,0 +1,33 @@
...
+            @NotNull Map<String, String> etiologyPerSignature)
+    {
+        if(signatureAllocation == null)
+        {
+            return null;
+        }
+        return signatureAllocation.stream()
+                .map(signature -> ImmutableSignatureAllocation.builder()
+                        .signature(signature.signature())
+                        .etiology(etiologyPerSignature.getOrDefault(signature.signature(), "-"))
```
kduyvesteyn, line 27: I would think this is fine, but you may also want to consider adding some extra safeties. Imagine Sigs changes in the meantime and suddenly writes "sig1" rather than "Sig1". Would you want to get "-" etiology without being aware? 

You could also consider mapping MISALLOC to "-", and raise a warning in case any other signature is missing from the map? 

I don't have a super strong opinion though, could go either way

orange/src/main/java/com/hartwig/hmftools/orange/algo/sigs/SigsInterpreter.java
```diff
@@ -0,0 +1,33 @@
...
+import com.hartwig.hmftools.datamodel.sigs.ImmutableSignatureAllocation;
+import com.hartwig.hmftools.datamodel.sigs.SignatureAllocation;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class SigsInterpreter
+{
+    @Nullable
+    public static List<SignatureAllocation> convertAndAnnotateWithEtiology(
```
kduyvesteyn, line 16: I would rename function to `interpret' to be more compatible with e.g. PurpleInterpreter

orange/src/main/java/com/hartwig/hmftools/orange/report/tables/SignatureAllocationTable.java
```diff
@@ -33,13 +33,15 @@ public static Table build(@NotNull String title, float width, @NotNull List<Sign
 
         Cells cells = new Cells(reportResources);
         Table table = Tables.createContent(width,
-                new float[] { 1, 1, 1, 3 },
-                new Cell[] { cells.createHeader("Signature"), cells.createHeader("Allocation"), cells.createHeader("Percent"),
+                new float[] { 1, 2.3F, 1, 1, 1 },
```
kduyvesteyn, line 36: One note about these numbers: They are all ratios, and I typically don't use floating point numbers.

If you multiple every number by 10 you get the same table (10, 23, 10, 10, 10) which I think would be a bit cleaner (but its a matter of taste!)

orange/src/test/java/com/hartwig/hmftools/orange/algo/sigs/SigsInterpreterTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+    static
+    {
+        etiologyPerSignature.put("Sig1", "APOBEC");
+    }
+
+    @Test
+    public void shouldConvertAndAnnotateSignatures()
+    {
+        assertNotNull(SigsInterpreter.convertAndAnnotateWithEtiology(signature, etiologyPerSignature));
```
kduyvesteyn, line 29: You could potentially expand on this test? E.g. assert that you do end up with APOBEC as signature for example?

Also, you could test the scenario where `null` is fed into the interpreter?

kduyvesteyn (approved):
Looks good!

Have a bunch of smaller comments, feel free to followup but also fine if you ignore!

PR #586, opened 2024-07-15
ACTIN-50: Show all viable fusions in ORANGE in case no HIGH driver events

cbruel:


orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/linx/LinxRecord.java
```diff
@@ -30,6 +30,9 @@ public interface LinxRecord
     @NotNull
     List<LinxFusion> additionalSuspectSomaticFusions();
 
+    @NotNull
+    List<LinxFusion> additionalViableFusions();
```
kduyvesteyn, line 34: I would recommend to rename to `additionalViableSomaticFusions()` to make it clear it is somatic only (in theory there could be germline fusions as well though not sure in practice... but in any case every property in linx currently has somatic or germline in their name

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/SomaticFindingsChapter.java
```diff
@@ -223,12 +226,14 @@ private void addFusions(@NotNull Document document)
     {
         String driverFusionsTitle = "Driver fusions";
         String nonDriverFusionsTitle = "Other potentially interesting fusions";
+        String inFrameFusionsTitle = "Potentially interesting in-frame fusions in case no high drivers detected";
```
kduyvesteyn, line 229: I would make the naming more clear. Previously we had "drivers" and "non-drivers" but now there is a 3rd category (which are also non-driver!)

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/SomaticFindingsChapter.java
```diff
@@ -395,4 +415,26 @@ private static <T> List<T> max10(@NotNull List<T> elements)
     {
         return elements.subList(0, Math.min(10, elements.size()));
     }
+
+    private static boolean hasHighDriverEvents(@NotNull List<com.hartwig.hmftools.datamodel.linx.LinxFusion> somaticFusions,
```
kduyvesteyn, line 419: I don't think you need the full package here for LinxFusion?

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/DnaFusionSelectorTest.java
```diff
@@ -97,6 +98,30 @@ public void canSelectFusionOfOncogene()
...
+        LinxFusion inFrameFusion = LinxTestFactory.fusionBuilder()
+                .phased(FusionPhasedType.INFRAME)
+                .chainTerminated(false)
+                .name("in frame fusion")
+                .build();
+        LinxFusion outOfFrameFusion = LinxTestFactory.fusionBuilder()
+                .phased(FusionPhasedType.OUT_OF_FRAME)
+                .chainTerminated(false)
+                .name("out of frame fusion")
+                .build();
```
kduyvesteyn, line 113: Fairly minor but your test relies on `fusionBuilder` to have reported=false as default. It may make your test more readable and resilient to explicitly set this parameter. Also that would be consistent with the other tests in this class.

kduyvesteyn (commented):
Looks good! 

Just a few minor comments, plus have reformatted some code that somehow wasn't auto-reformatted properly (see my latest commit)

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/SomaticFindingsChapter.java
cbruel: Yes you're right, happened because I moved this function (first it used the other `LinxFusion`)

orange/README.md
```diff
@@ -148,6 +148,8 @@ interesting and added to the report:
 - Other potentially relevant fusions. A maximum of 10 additional fusions (picked arbitrarily) are reported as potentially interesting:
     1. Any fusion that is not reported and has a reported type other than NONE.
     2. Any fusion in a gene that is configured as an oncogene in the driver gene panel.
+- Other potentially interesting in-frame fusions in case no high drivers events are detected
+    1. In case no high driver events are detected, any in frame non chain terminated fusion that is not already reported
```
kduyvesteyn, line 152: Super minor but I think it should be `inframe` or `in-frame`, not 100% sure!

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/DnaFusionSelector.java
```diff
@@ -34,6 +35,16 @@ public static List<LinxFusion> selectInterestingUnreportedFusions(@NotNull List<
         return filtered;
     }
 
+    @NotNull
+    public static List<LinxFusion> selectViableSomaticFusions(@NotNull List<LinxFusion> somaticFusions,
```
kduyvesteyn, line 39: Another minor but maybe add "Additional" to function name? The function name now implies you select all viable fusions, while you only select viable fusions that are not otherwise reported.

kduyvesteyn (approved):
Looks good!! 2 minor comments

PR #588, opened 2024-07-16
ACTIN-293: Ensure HIV is never reported in ORANGE report

cbruel:


hmf-common/src/main/java/com/hartwig/hmftools/common/virus/TaxidType.java
kduyvesteyn, line 1: Could this class be moved to virus interpreter module? Or do you have a reason why you put it in common module?

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/algo/VirusBlacklistingDb.java
```diff
@@ -0,0 +1,26 @@
...
+import org.jetbrains.annotations.Nullable;
+
+@Value.Immutable
+@Value.Style(allParameters = true,
+             passAnnotations = { NotNull.class, Nullable.class })
+public abstract class VirusBlacklistingDb
+{
+
+    @NotNull
+    public abstract Integer taxid();
```
kduyvesteyn, line 16: This should be `int` :)

(your favourite kotlin vs java issue)

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/algo/VirusBlacklistingDb.java
kduyvesteyn, line 1: Wouldn't `VirusBlacklistingEntry` be a better name for this class?

Also, it doesn't seem like `type`, `name` and `reason` are used?

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/VirusInterpreterAlgo.java
```diff
@@ -104,6 +109,21 @@ boolean report(VirusBreakend virusBreakend, double expectedClonalCoverage, final
...
     }
 
+    @VisibleForTesting
+    boolean blacklist(int taxidGenus, int taxidSpecies, int taxid, boolean reported)
+    {
+        boolean blacklisted = virusBlacklistingDb.stream()
+                .anyMatch(blacklistedVirus -> taxidGenus == blacklistedVirus.taxid() || taxidSpecies == blacklistedVirus.taxid());
+        if(blacklisted && reported)
+        {
+            throw new RuntimeException("Virus with taxid {} configured as reported and as blacklisted" + taxid);
```
kduyvesteyn, line 119: I like the exception here! Though it is more typical to throw an `IllegalStateException`

Also, may be an idea to add this check one level higher? In that case you don't have to pass the actual taxid into this function?

orange/src/main/java/com/hartwig/hmftools/orange/algo/virus/VirusInterpreter.java
```diff
@@ -0,0 +1,17 @@
+package com.hartwig.hmftools.orange.algo.virus;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.hartwig.hmftools.common.virus.AnnotatedVirus;
+
+import org.jetbrains.annotations.NotNull;
+
+public class VirusInterpreter
```
kduyvesteyn, line 10: A class with only static methods should be declared `final` to avoid inheritance

orange/src/main/java/com/hartwig/hmftools/orange/algo/virus/VirusInterpreter.java
```diff
@@ -0,0 +1,17 @@
...
+{
+    @NotNull
+    public static List<AnnotatedVirus> filterOutBlacklistedViruses(@NotNull List<AnnotatedVirus> allViruses)
```
kduyvesteyn, line 13: I think I would rename to just `filterBlacklisted...`

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/virus/VirusInterpreterEntry.java
```diff
@@ -30,6 +30,8 @@ public interface VirusInterpreterEntry
 
     boolean reported();
 
+    boolean blacklisted();
+
```
kduyvesteyn, line 34: Why did you decide to add this property if you remove blacklisted virus entries anyways? 

In other words, the only producer of ORANGE datamodel guarantees this property is false - maybe cleaner to just remove it?

kduyvesteyn (commented):
Few things in code that I didn't quite understand!

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/algo/VirusBlacklistingDb.java
cbruel: I will put this above my bed, so I can't forget anymore :)

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/virus/VirusInterpreterEntry.java
cbruel: Yes you are totally correct, doesn't make sense!

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/VirusInterpreterAlgo.java
cbruel: Thanks, much better indeed!

hmf-common/src/main/java/com/hartwig/hmftools/common/virus/TaxidType.java
cbruel: I put it there since all other virus interpreter datamodel classes where there (`AnnotatedVirus`, `VirusType` etc.). But this class isn't used in any other module, so I moved it to `virus-interpreter` module.

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/algo/VirusBlacklistingDb.java
cbruel: @kduyvesteyn Yes I had doubts about this part. You're correct that the other fields aren't used, but I thought it might be cleaner to save all the fields from the tsv, if we might want to use it in the future. But `VirusBlacklistingDbFile.loadFromTsv(`) could actually just output a list of int's (no need for a class at all). Or do you prefer a `VirusBlackListingEntry` with 1 `taxid()` field? Seems superfluous to me to have a class with only 1 datafield?

hmf-common/src/main/java/com/hartwig/hmftools/common/virus/TaxidType.java
kduyvesteyn: It would indeed not be "wrong" to keep it in common for the reasons you mention, but my preference typically is to keep common as small as possible (it basically contains the datamodel that the different tools need to communicate with each other)

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/algo/VirusBlacklistingDb.java
kduyvesteyn: Ah I get it. I would just only read the taxid then, and store it in a list of integers. I wouldn't create a class with 1 int field, feels superfluous indeed. 

It's not unusual we keep extra fields for (human) context in TSVs that we then don't read because algo's don't need them.

hmf-common/src/main/java/com/hartwig/hmftools/common/virus/TaxidType.java
kduyvesteyn: Related: If you replace the VirusBlacklistingDb with an integer then you don't need this enum anymore

orange/README.md
```diff
@@ -265,6 +265,7 @@ investigate potential causes for QC failure.
   - Add etiology information to signatures and sort by allocation
   - Add percentage of unsupported segments to Quality control page
   - Show all viable fusions in ORANGE in samples where we detect no HIGH drivers
+  - Ensure HIV is never reported in ORANGE report
```
kduyvesteyn, line 268: Maybe add "or included in ORANGE json"?

orange/src/main/java/com/hartwig/hmftools/orange/conversion/OrangeConversion.java
```diff
@@ -101,8 +102,8 @@ public static LilacAllele convert(@NotNull com.hartwig.hmftools.common.hla.Lilac
     public static VirusInterpreterData convert(@NotNull com.hartwig.hmftools.common.virus.VirusInterpreterData interpreterData)
     {
         return ImmutableVirusInterpreterData.builder()
-                .allViruses(ConversionUtil.mapToIterable(interpreterData.allViruses(), OrangeConversion::convert))
-                .reportableViruses(ConversionUtil.mapToIterable(interpreterData.reportableViruses(), OrangeConversion::convert))
+                .allViruses(ConversionUtil.mapToIterable(VirusInterpreter.filterBlacklistedViruses(interpreterData.allViruses()), OrangeConversion::convert))
+                .reportableViruses(ConversionUtil.mapToIterable(VirusInterpreter.filterBlacklistedViruses(interpreterData.reportableViruses()), OrangeConversion::convert))
```
kduyvesteyn, line 106: Minor and not 100% sure but you could again make a "VirusInterpreterAlgo" that filters the blacklisted viruses after loading, and before conversion, so we can keep the conversion code trivial (no actual logic)

orange/src/test/java/com/hartwig/hmftools/orange/algo/virus/VirusInterpreterTest.java
```diff
@@ -0,0 +1,44 @@
...
+    public void canFilterBlacklistedViruses()
+    {
+        List<AnnotatedVirus> filteredViruses =
+                VirusInterpreter.filterBlacklistedViruses(Lists.newArrayList(createAnnotatedVirus("blacklisted virus", false, true), createAnnotatedVirus("non blacklisted virus", true, false)));
+        assertEquals(1, filteredViruses.size());
+        assertEquals(filteredViruses.iterator().next().name(), "non blacklisted virus");
+    }
+
+    @NotNull
+    public static AnnotatedVirus createAnnotatedVirus(@NotNull String name, boolean reported, boolean blacklisted)
```
kduyvesteyn, line 28: I suspect this function can be private?

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/VirusInterpreterAlgo.java
```diff
@@ -104,6 +115,13 @@ boolean report(VirusBreakend virusBreakend, double expectedClonalCoverage, final
         return reported && virusQCStatus;
     }
 
+    @VisibleForTesting
+    boolean blacklist(int taxidGenus, int taxidSpecies)
+    {
+        return blacklistedTaxids.stream()
```
kduyvesteyn, line 121: You could also use `contains` perhaps with a simple list of integers? 

No super strong opinion but think it may be a bit more readable that way !(blacklisted.contains(genus) || blacklisted.contains(species))

virus-interpreter/src/main/java/com/hartwig/hmftools/virusinterpreter/VirusInterpreterConfig.java
```diff
@@ -44,7 +47,8 @@ public VirusInterpreterConfig(final ConfigBuilder configBuilder)
         PurpleDir = configBuilder.getValue(PURPLE_DIR_CFG);
         VirusBreakendTsv = configBuilder.getValue(VIRUS_BREAKEND_TSV);
         TaxonomyDbTsv = configBuilder.getValue(TAXONOMY_DB_TSV);
-        VirusReportedDbTsv = configBuilder.getValue(VIRUS_REPORTING_DB_TSV);
+        VirusReportedDbTsv = configBuilder.getValue(VIRUS_BLACKLISTING_DB_TSV);
```
kduyvesteyn, line 50: Do you read the blacklisting TSV twice? Looks like an "oops" moment? :)

kduyvesteyn (approved):
Looks good!

A few minor comments and also I think one bug.

orange/src/main/java/com/hartwig/hmftools/orange/conversion/OrangeConversion.java
cbruel: Moved it

PR #591, opened 2024-07-18
INFRA-541 Orange Dockerfile, call it from build

nedleitch:


orange/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openjdk:17-alpine
```
pauldwolfe, line 1: Are we sure that hmf-tools run on 17? I think we normally extend from `eclipse-temurin:11-jre`

pauldwolfe (commented):
Looks like a good, simple start! I guess the docker push step is intentionally left out for now?

nedleitch:
Closing this PR in favour of the one in branch INFRA-541-3 as that one starts with an Orange version that works properly.

The changes from here have just been copied into there manually.

PR #596, opened 2024-07-23
INFRA-541 Build Orange Docker container

nedleitch:
Add the creation of a Docker container to the building of `hmftools` modules. Does not do the Docker part if there is no Dockerfile for the module.

I've leveraged the Python script to do the heavy lifting but since the containers used in the build environment are quite modular I opted to generate the Docker-related commands for execution by a subsequent stage. I feel given the simplicity of the build overall this is acceptable.

### DIFF
--- a/hmftools-build.py
+++ b/hmftools-build.py
@@ -103,7 +103,7 @@ class Release:
         base_url="{}/{}/assets?name".format(self._construct_url("uploads"), id)
         response = requests.post(f"{base_url}={self.module}_v{self.version}.jar", 
                 headers = headers, 
-                data = this.artifact_file.read())
+                data = self.artifact_file.read())
         response.raise_for_status()
         print(f"Uploaded {artifact.name}")
 
--- a/hmftools-build.py
+++ b/hmftools-build.py
@@ -70,7 +70,7 @@ class Release:
         self._upload_artifacts(id)
 
     def _create_release(self):
-        print(f"Creating release {self.release_name}")
+        print(f"Creating release [{self.release_name}]")
         request = {"tag_name": self.release_name,
                 "target_commitish": "master",
                 "name": self.release_name,
@@ -86,6 +86,7 @@ class Release:
         response = requests.post(self._construct_url("api"),
                 json = request,
                 headers = headers)
+        print(f"Response: {response.text}")
         response.raise_for_status()
         return response.json()["id"]
 
@@ -166,7 +167,7 @@ def build_and_release(raw_tag: str):
     Maven.deploy_all(module_pom, *dependencies_pom)
 
     Docker(module, version).build()
-    Release(f"{module} v{tag}", [open(f"/workspace/{module}/target/{module}-{version}-jar-with-dependencies.jar", "rb")], 
+    Release(f"{module} v{version}", [open(f"/workspace/{module}/target/{module}-{version}-jar-with-dependencies.jar", "rb")], 
             open("/workspace/github.token", "r").read()).create()
 
 if __name__ == '__main__':
--- a/hmftools-build.py
+++ b/hmftools-build.py
@@ -16,6 +16,8 @@ Example:
 """
 import re
 import subprocess
+import requests
+import os.path
 from xml.etree import ElementTree
 from argparse import ArgumentParser
 
--- a/hmftools-build.py
+++ b/hmftools-build.py
@@ -59,11 +59,11 @@ class Docker:
 
 
 class Release:
-    def __init__(self, tag_name, module, version, artifact_files, token):
+    def __init__(self, tag_name, module, version, artifact_file, token):
         self.tag_name = tag_name
         self.module = module
         self.version = version
-        self.artifact_files = artifact_files
+        self.artifact_file = artifact_file
         self.token = token
         self.release_name = f"{module} v{version}"
 
@@ -94,20 +94,16 @@ class Release:
         return response.json()["id"]
 
     def _upload_artifacts(self, id):
-        print(f"Uploading artifacts to release {id}")
+        print(f"Uploading artifact to release {id}")
         headers = {"Accept": "application/vnd.github+json",
                 "Authorization": f"Bearer {self.token}",
                 "X-GitHub-Api-Version": "2022-11-28",
                 "Content-Type": "application/octet-stream"
         }
         base_url="{}/{}/assets?name".format(self._construct_url("uploads"), id)
-        for artifact in self.artifact_files:
-            response = requests.post(f"{base_url}={self.module}_v{self.version}.jar",
-                    headers = headers,
-                    data = artifact.read())
-            response.raise_for_status()
-            print(f"Uploaded {artifact.name}")
-        print("Artifacts uploaded")
+        response = requests.post(f"{base_url}={self.module}_v{self.version}.jar", headers = headers, data = artifact_file.read())
+        response.raise_for_status()
+        print(f"Uploaded {artifact.name}")
 
     def _construct_url(self, prefix):
         return f"https://{prefix}.github.com/repos/hartwigmedical/hmftools/releases"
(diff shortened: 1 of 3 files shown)


hmftools-build.py
```diff
@@ -39,9 +39,22 @@ def set_version(self, version):
...
+        #subprocess.run(['mvn', 'deploy', '-B', '-pl', module_str, '-am', '-DdeployAtEnd=true'], check=True)
+        subprocess.run(['mvn', 'install', '-B', '-pl', module_str, '-am'], check=True)
 
 
+class Docker:
+    def __init__(self, module, version):
+        self.module = module
+        self.version = version
+        self.internal_image = f'europe-west4-docker.pkg.dev/hmf-build/hmf-docker/{self.module}:{self.version}'
+        self.external_image = f'europe-west4-docker.pkg.dev/hmf-build/build-registry-docker/{self.module}:{self.version}'
```
mvanniekerkHartwig, line 51: So if I understand correctly, in your approach you're choosing to release to an open-to-the-world artifact registry? Did you also consider releasing to docker central? (I think this approach is fine, but it could be good to also consider releasing to docker central for visibility.)

mvanniekerkHartwig (commented):
Code looks good, I think it will work just fine! I have 2 concerns though, that you probably thought about as well, so I'm curious to hear your thoughts:

1. To make this project succeed, we should also include tools (and tool versions) that were released in the past (maybe a long time ago). How are you planning to deal with those, since you are working from the master branch of hmftools? With this approach it will be tough to "go back in history".
2. Not everyone that releases tools in this repo will stick to "our" tag-release style. Also, there is no guarantee that everyone that releases tools on this repo will include the _same_ jar in the [releases page](https://github.com/hartwigmedical/hmftools/releases) that you get from running `mvn clean install` on the commit that was tagged. I believe there have been (please correct me if I'm wrong) multiple cases in the past where the commit that was tagged with the release does not compile or does not pass tests, and the jar that was included in the releases page was actually generated from a different commit, or from a code state that was never committed. Therefore to work this approach may require some behaviour changes that may or may not be feasible to ask from all contributors on this project.

Because of points 1 and 2, I advocated an approach of putting the docker files in a separate repo, and adding some "subscription" logic, that creates a new docker image every time a new jar is added to the [releases page](https://github.com/hartwigmedical/hmftools/releases). Did you also consider this approach? I'd like to hear your thoughts on this, how you are going to address point 1, and whether you think point 2 is a real problem or can be ignored.

hmftools-build.py
nedleitch: Haven't quite pinned down the external repo, I'd be just fine with Docker central as well, but the principal requirement here was that Stephen (Watts, in Melbourne, who has built the Nextflow implementation) would be able to access them. I kicked the can down the road a bit thinking I'd just put this in and deal with the details of exactly where it is going to go once we've got everybody on board, but maybe I'll give it a bit more thought up-front to bolster the case for this approach.

nedleitch:
> Code looks good, I think it will work just fine! I have 2 concerns though, that you probably thought about as well, so I'm curious to hear your thoughts:
>
Yeah had the same thoughts and decided to go with this approach because I felt that if it were easy enough there would be no reason not to use it, and then maybe we could just all get what we want. While that's an idealistic approach I don't think it is naively so because doing it this way saves so much work for everybody involved. I agree with your points but also thought it might be worth trying to build it the way we would like it to work first, then seeing if we can get some consensus rather than assuming we can't improve the process.

> 1. To make this project succeed, we should also include tools (and tool versions) that were released in the past (maybe a long time ago). How are you planning to deal with those, since you are working from the master branch of hmftools? With this approach it will be tough to "go back in history".

I figured we would want some of the old releases indeed. In my mind those would all be constructed from the existing Github releases. I had a fair bit of interaction with the Github APIs while I was writing this and am pretty confident I can put something together to scrape all the old releases, pluck out the JARs that were published and make Docker containers/publish Maven artifacts for them. 

> 2. Not everyone that releases tools in this repo will stick to "our" tag-release style. Also, there is no guarantee that everyone that releases tools on this repo will include the _same_ jar in the [releases page](https://github.com/hartwigmedical/hmftools/releases) that you get from running `mvn clean install` on the commit that was tagged. I believe there have been (please correct me if I'm wrong) multiple cases in the past where the commit that was tagged with the release does not compile or does not pass tests, and the jar that was included in the releases page was actually generated from a different commit, or from a code state that was never committed. Therefore to work this approach may require some behaviour changes that may or may not be feasible to ask from all contributors on this project.
> 

There have definitely been cases of this and it caused lots of headaches. The hours I spent generating md5s from buckets and images I will never get back! I think this is exactly why it's important to try to put down tooling that aligns with how we want to work rather than try to deal with exceptions that deviate from the norm. But this has been essentially my only hangup about doing it this way. I go back to what I said above which is I'd rather build something that aligns with how we want to work than try to accommodate a suboptimal de facto process.

> Because of points 1 and 2, I advocated an approach of putting the docker files in a separate repo, and adding some "subscription" logic, that creates a new docker image every time a new jar is added to the [releases page](https://github.com/hartwigmedical/hmftools/releases). Did you also consider this approach?

I didn't actually. Interesting idea. One thing about that is the releases are really quite mutable so it might be difficult to enforce a level of consistency that would probably be needed to get this right (artifact naming and uniqueness, release naming, etc).

hmftools-build.py
pauldwolfe: I think we'll start with pushing internally and to DockerHub. DockerHub will be accessible to NextFlow. I guess the "external" image above is just a placeholder?

Bioconda, which is currently used by Stephen requires a separate "recipe" and approval process for each change, so let's consider it a non-starter for now.

nedleitch:
I got a passing build that included the push to `docker.io`, so the image is now available on Dockerhub, but the repository is currently private. We will need to manually change the repository to public when we are ready.

But open question on this: the Docker container includes the JAR with all the dependencies, and it's not clear to me whether we might run afoul of legal issues if we publish this openly due to all the libraries we include.

Have been down this road in the past and I think it essentially comes down to having to verify everything has a permissive-enough licence. Any thoughts from either of you? Paul do you remember doing anything on this front before publishing any of the other public containers?

hmftools-build.py
```diff
@@ -39,7 +41,76 @@ def set_version(self, version):
     @staticmethod
     def deploy_all(*modules):
         module_str = ','.join([m.name for m in modules])
-        subprocess.run(['mvn', 'deploy', '-B', '-pl', module_str, '-am', '-DdeployAtEnd=true'], check=True)
+        #subprocess.run(['mvn', 'deploy', '-B', '-pl', module_str, '-am', '-DdeployAtEnd=true'], check=True)
```
pauldwolfe, line 44: We'd want to uncomment this before merging?

hmftools-build.py
```diff
@@ -39,7 +41,76 @@ def set_version(self, version):
...
+        subprocess.run(['mvn', 'install', '-B', '-pl', module_str, '-am'], check=True)
+
+
+class Docker:
+    def __init__(self, module, version):
+        self.module = module
+        self.version = version
+        self.private_image = f'europe-west4-docker.pkg.dev/hmf-build/hmf-docker/{self.module}:{self.version}'
+        self.public_image = f'hartwigmedicalfoundation/{self.module}:{self.version}'
```
pauldwolfe, line 53: private and public are terms we used to indicate one image has "private" stuff in it, while in this case its really the same image. I liked internal and external better.

hmftools-build.py
```diff
@@ -39,7 +41,76 @@ def set_version(self, version):
...
+
+    def build(self):
+        with open("/workspace/docker.sh", "w") as output:
+            output.write(f'docker build {self.module} -t {self.private_image} -t {self.public_image} --build-arg VERSION={self.version}\n')
+            output.write(f'docker push {self.private_image}\n')
+            output.write(f'docker login -u hartwigmedicalfoundation -p $(cat /workspace/dockerhub.password)\n')
+            output.write(f'docker push {self.public_image}\n')
+
+
+class Release:
```
pauldwolfe, line 63: Maybe GithubRelease so it's clear what this is doing?

pauldwolfe (approved):

nedleitch:
> I got a passing build that included the push to `docker.io`, so the image is now available on Dockerhub, but the repository is currently private. We will need to manually change the repository to public when we are ready.
> 
> But open question on this: the Docker container includes the JAR with all the dependencies, and it's not clear to me whether we might run afoul of legal issues if we publish this openly due to all the libraries we include.
> 
> Have been down this road in the past and I think it essentially comes down to having to verify everything has a permissive-enough licence. Any thoughts from either of you? Paul do you remember doing anything on this front before publishing any of the other public containers?

I spoke with Paul about this and he confirmed these things had all been done before we started releasing the "with dependencies" JARs via Github releases so there should be nothing more to do here.

PR #597, opened 2024-07-29
ACTIN-824: Add "Presence of tumor stats" to quality control page

kzuberihmf:
Adds tumorStats to orange-datamodel under purple, and corresponding code to populate in orange.

The orange pdf will have the tumor stats in a new table in the quality control chapter.

One of the attributes required the purple segments.tsv file which is not currently loaded, and that doesn't have a loader in common. I added support to extract just the minimal fields required to orange.


ninajacobs:
See some comments in Slack :)

kzuberihmf:
The purple constants don't seem to be available in common, so I tried the minimal version of the table without the thresholds listed. Also edited for your suggested wording, below is the updated table on COLO. Happy to tweak further as needed.

For the convert-germline-to-somatic case I've added in the germline hotspot mutations to the calculation. And the tumor max diploid proportion is included, extracted from the purple purity.tsv. Also updated the comparison scripts in data-vm experiments for these changes.

Screenshot 2024-08-01 at 8 51 21 AM

ninajacobs:
Hi @kzuberihmf, nice work thanks, will be a helpful table! 

I'd propose to still add the subtitle: "At least one requirement needs to be met to indicate detection of tumor, in case of highly diploid candidate solutions (proportion >=0.97)" in which ideally 0.97 should be derived from "HIGHLY_DIPLOID_PERCENTAGE_DEFAULT" in PurpleConstants to make it future-safe?

And maybe make the table content grey in case tumor maximum diploid proportion < HIGHLY_DIPLOID_PERCENTAGE_DEFAULT ?

And I'd propose to change the title to "Tumor detection requirements" after all.

Finally, I'd say 'proportion' is a value like 0.02 instead of 2% (so propose to use 0.02).

orange-datamodel/src/test/resources/minimally.populated.orange.json
```diff
@@ -316,7 +316,15 @@
         "maxCopyNumber": 1.2,
         "minMinorAlleleCopyNumber": 0.4
       }
-    ]
+    ],
+    "tumorStats": {
```
cbruel, line 320: I don't think these should be (all) 0? There is for example a hotspot mutation in this json.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return purpleData.allSomaticVariants().stream()
+                .filter(variant -> variant.type() == VariantType.SNP)
+                .mapToInt(variant -> variant.allelicDepth().AlleleReadCount)
+                .sum();
+    }
+
+    private static int hotspotMutationCount(@NotNull PurpleData purpleData, boolean convertGermlineToSomatic)
+    {
+
```
cbruel, line 57: Remove white space?

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,91 @@
...
+                    .filter(variant -> variant.tier() == VariantTier.HOTSPOT)
+                    .count();
+        }
+
+        return count;
+    }
+
+    private static int bafCount(@NotNull PurpleData purpleData)
+    {
+
```
cbruel, line 74: Remove white space?

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/QualityControlChapter.java
```diff
@@ -256,4 +257,36 @@ private void addSageBQRPlots(@NotNull Document document)
...
+        tumorStats.addCell(cells.createContent(String.valueOf(report.purple().tumorStats().smallVariantAlleleReadCount())));
+
+        tumorStats.addCell(cells.createContent("Sum of structural variant tumor fragment counts (excluding single breakends)"));
+        tumorStats.addCell(cells.createContent(String.valueOf(report.purple().tumorStats().structuralVariantTumorFragmentCount())));
+
+        tumorStats.addCell(cells.createContent("Sum of B-allele frequency points in germline diploid regions regions with tumor ratio < 0.8 OR > 1.2"));
+        tumorStats.addCell(cells.createContent(String.valueOf(report.purple().tumorStats().bafCount())));
+
+        document.add(new Tables(reportResources).createWrapping(tumorStats,
+                "Detection of tumor"));
```
cbruel, line 290: If we are going to add a subtitle “At least one requirement needs to be met .. “, I think we should add the requirements as well to the table. Maybe you could add these between parentheses like was written in the jira ticket: “Number of hotspot mutations (should be > 0)”.

cbruel (commented):
Some other things I learned from Korneel to keep in mind to not forget :)
- Updating the README (Making an upcoming section)
- Update the COLO829 test reports by running ReportTestGeneratorApplication (WGS = Tumor_only=false, Targeted = Tumor_only=true)
- You could also use that script to update real.orange.json (see comment in script). That way the “samplingDate” is adjusted to the current date.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private static int hotspotMutationCount(@NotNull PurpleData purpleData, boolean convertGermlineToSomatic)
+    {
+
+        int count = (int) purpleData.allSomaticVariants().stream()
+                .filter(variant -> variant.tier() == VariantTier.HOTSPOT)
+                .count();
+
+        if(convertGermlineToSomatic && purpleData.allGermlineVariants() != null)
```
ninajacobs, line 62: I think if this is 'reportableGermlineVariants' it solves our problem we found :)

kzuberihmf: Thanks. To confirm, for small variant allele read counts I'm still using allSomaticVariants() without checking reportable, should this  also be reportableSomaticVariants()?

ninajacobs: Yes for that one it should be all passing SNPs (so not necessarily reportable indeed). David will review this though still too.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,91 @@
...
+        return purpleData.allSomaticVariants().stream()
+                .filter(variant -> variant.type() == VariantType.SNP)
+                .mapToInt(variant -> variant.allelicDepth().AlleleReadCount)
+                .sum();
+    }
+
+    private static int hotspotMutationCount(@NotNull PurpleData purpleData, boolean convertGermlineToSomatic)
+    {
+
+        int count = (int) purpleData.allSomaticVariants().stream()
```
ninajacobs, line 58: @kzuberihmf just to be sure; is allSomaticVariants already filtered on filter=PASS?

kzuberihmf: Yes its is loaded using `PurpleVariantContextLoader.withPassingOnlyFilter()` in PurpleDataLoader

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/QualityControlChapter.java
kzuberihmf: Yes we did have something like that in an earlier version, I took it out because I had to hardcode the thresholds to be consistent with Purple. But agree it would easier for someone to interpret with those thresholds displayed. Guess I can add them back.

kzuberihmf:
> Some other things I learned from Korneel to keep in mind to not forget :)
> 
> * Updating the README (Making an upcoming section)
> * Update the COLO829 test reports by running ReportTestGeneratorApplication (WGS = Tumor_only=false, Targeted = Tumor_only=true)
> * You could also use that script to update real.orange.json (see comment in script). That way the “samplingDate” is adjusted to the current date.

Thanks, I've updated these. Though we'll likely update the report layout a bit more.

kzuberihmf:
> 
> I'd propose to still add the subtitle: "At least one requirement needs to be met to indicate detection of tumor, in case of highly diploid candidate solutions (proportion >=0.97)" in which ideally 0.97 should be derived from "HIGHLY_DIPLOID_PERCENTAGE_DEFAULT" in PurpleConstants to make it future-safe?
> 
> And maybe make the table content grey in case tumor maximum diploid proportion < HIGHLY_DIPLOID_PERCENTAGE_DEFAULT ?
> 

Added this, with the value hard-coded. I'm not sure how to do subtitles so for the moment its on a separate line but in the same font as the heading. 

> And I'd propose to change the title to "Tumor detection requirements" after all.
> 
> Finally, I'd say 'proportion' is a value like 0.02 instead of 2% (so propose to use 0.02).

Updated.

cbruel:
> > I'd propose to still add the subtitle: "At least one requirement needs to be met to indicate detection of tumor, in case of highly diploid candidate solutions (proportion >=0.97)" in which ideally 0.97 should be derived from "HIGHLY_DIPLOID_PERCENTAGE_DEFAULT" in PurpleConstants to make it future-safe?
> > And maybe make the table content grey in case tumor maximum diploid proportion < HIGHLY_DIPLOID_PERCENTAGE_DEFAULT ?
> 
> Added this, with the value hard-coded. I'm not sure how to do subtitles (or grey tables) so for the moment its on a separate line but in the same font as the heading.
> 
> > And I'd propose to change the title to "Tumor detection requirements" after all.
> > Finally, I'd say 'proportion' is a value like 0.02 instead of 2% (so propose to use 0.02).
> 
> Updated.

I pushed a commit which makes the subtitle and reduces the distance between the title and table a bit, because it was quite large :)

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/QualityControlChapter.java
```diff
@@ -256,4 +259,31 @@ private void addSageBQRPlots(@NotNull Document document)
...
+
+        tumorStats.addCell(cells.createContent("Sum of structural variant tumor fragment counts (excluding single breakends)"));
+        tumorStats.addCell(cells.createContent(String.valueOf(report.purple().tumorStats().structuralVariantTumorFragmentCount())));
+
+        tumorStats.addCell(cells.createContent("Sum of B-allele frequency points in germline diploid regions regions with tumor ratio < 0.8 OR > 1.2"));
+        tumorStats.addCell(cells.createContent(String.valueOf(report.purple().tumorStats().bafCount())));
+
+        document.add(new Tables(reportResources).createWrapping(tumorStats,
+                "Tumor detection requirements", "At least one requirement needs to be met to indicate detection of tumor, in case of highly diploid candidate solutions (proportion >=0.97)\n\n"));
+    }
```
DFKoetsier, line 288: We might want to be careful with this subtitle for two reasons:
- In targeted and/or tumor-only mode the FAIL_NO_TUMOR check is completely different.
- Even when passing these requirements, a FAIL_NO_TUMOR can still be called if the somatic purity fit fails.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactoryTest.java
```diff
@@ -0,0 +1,241 @@
...
+        return ImmutableStructuralVariantImpl.builder()
+                .from(variant)
+                .start(ImmutableStructuralVariantLegImpl.builder().from(variant.start())
+                        .tumorVariantFragmentCount(count).build())
+                .build();
+    }
+
+    @NotNull
+    private static PurpleVariantContext purpleVariantContext(VariantType variantType, boolean reported,
+            int alleleReadCount)
```
DFKoetsier, line 197: Maybe unnecessary newline?

DFKoetsier (commented):
Looks good! I made some remarks and suggestions I'm not sure about, but nothing major.

kzuberihmf:
> 
> I pushed a commit which makes the subtitle and reduces the distance between the title and table a bit, because it was quite large :)

Great, thanks! I was afraid to dig into figuring out how pdf formatting worked 😄

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+        return enrichedVariants.stream()
+                .filter(variant -> !variant.isFiltered() && variant.type() != StructuralVariantType.SGL)
+                .mapToInt(variant ->
+                {
+                    Integer count = variant.start().tumorVariantFragmentCount();
+                    return count != null ? count : 0;
+                })
+                .sum();
+    }
```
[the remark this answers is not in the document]
kzuberihmf: Ok updated, though still differs in the SGL check.  Also removes the EnrichedStructuralVariant conversion.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,89 @@
...
+        int count = (int) purpleData.allSomaticVariants().stream()
+                .filter(variant -> variant.tier() == VariantTier.HOTSPOT)
+                .count();
+
+        if(convertGermlineToSomatic && purpleData.reportableGermlineVariants() != null)
+        {
+            count += (int) purpleData.reportableGermlineVariants().stream()
+                    .filter(variant -> variant.tier() == VariantTier.HOTSPOT)
+                    .count();
+        }
```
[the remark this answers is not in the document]
kzuberihmf: This sounds like a nice cleanup. However at the point of germline conversion we would be operating on the PurpleRecord instead of the loaded PurpleData, and it doesn't appear that variant.tier() is included in the model at that point. Wouldn't this require updating the orange-datamodel to include tier and if so is that worth it?

We do have access to HotSpotType though, is it valid to switch to that?

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
```diff
@@ -0,0 +1,89 @@
...
+                .sum();
+    }
+
+    private static int smallVariantAlleleReadCount(@NotNull PurpleData purpleData)
+    {
+        return purpleData.allSomaticVariants().stream()
+                .filter(variant -> variant.type() == VariantType.SNP)
+                .mapToInt(variant -> variant.allelicDepth().AlleleReadCount)
+                .sum();
+    }
```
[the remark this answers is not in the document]
kzuberihmf: OK will wait for further clarification on this.

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/QualityControlChapter.java
kzuberihmf: So for the reporting section the remaining loose ends appear to be the wording of this, and if we should again include the count thresholds as mentioned earlier or if its ok to just display the observed counts. Welcome recommendations on this :smile:

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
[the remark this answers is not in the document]
DFKoetsier: It should be valid to switch to HotspotType, since `tier==HOTSPOT` and `hotspotType==HOTSPOT` are meant to be equivalent. 

HotspotType is determined by the `HOTSPOT` VCF tag, which is determined in Purple in `HotspotEnrichment` like this:
```
for(VariantHotspot hotspot : hotspots)
{
        if(variantStart == hotspot.position() && hotspot.ref().equals(variant.getReference().getBaseString())
        && variant.getAlternateAlleles().stream().map(Allele::getBaseString).collect(Collectors.toList()).contains(hotspot.alt()))
        {
            variant.getCommonInfo().putAttribute(HOTSPOT_FLAG, true);
            return;
        }
       ...
}
```
And the Hotspot TIER is set when the below is true in `HotspotSelector` in SAGE:
```
public boolean isHotspot(final SimpleVariant variant)
{
    final AtomicBoolean hotspotMatch = new AtomicBoolean(false);

    mHotspotSelector.select(variant.position(), variant.position(), hotspot ->
    {
        if(hotspot.alt().equals(variant.alt()) && hotspot.ref().equals(variant.ref()))
        {
            hotspotMatch.set(true);
        }
    });

    return hotspotMatch.get();
}
```
Where in this case the list of hotspots has already been filtered down to hotspots of the right chromosome.

So both just check whether there is a hotspot variant configured that matches in chromosome, position, ref and alt.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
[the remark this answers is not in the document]
DFKoetsier: As far as I understand the `variant.end() != null` check is also meant out to filter out the SGLs. But I like the extra clarity of excluding SGLs more explicitly, so looks good!

orange/src/main/java/com/hartwig/hmftools/orange/report/chapters/QualityControlChapter.java
ninajacobs: see mail!

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/TumorStatsFactory.java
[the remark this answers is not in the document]
ninajacobs: see mail!

kzuberihmf:
Thanks for the detailed feedback everyone, here and via email. I've updated to:

* revise title in report section, and remove subtitle
* omit report section in tumor-only mode
* perform the hotspot germline to somatic conversion in the GermlienConversion class
* updated sample pdf's

Also updated the test jar on data-vm experiments for testing verification. Hope I haven't missed anything 😄

ninajacobs:
I pushed a minor capital letter change to ACTIN-824. I also did some testing - no unexpectencies found (will Slack you about details).

cbruel:
I don't have any more comments :) (only removed the code for adding the subtitle, since it wasn't used anymore)

PR #602, opened 2024-08-20
INFRA-653 Switch to Github app authentication

nedleitch:
Rather than the PAT (personal access token) linked to my account, switch to using this flow meant for the Github "app" capability.

For testing I configured an app in Github and granted it the required permissions on one of my personal repositories. I generated a private key for it and used the client id and installation id metadata to test the flow implemented in the code.

In the code there is a change to read the private key rather than the PAT, and this has been parameterised instead of hardcoded. The client and installation ids are also passed in are used in conjunction with the private key to generate a JWT, which is then used to fetch a short-lived token. That token can then be used in the same way as the PAT.

I have also modified the cloudbuild file, a new secret will have to be made and the given parameters added to the build. Finally there is a separate PR for the build-tools container to add the required JWT library to the Python environment.

This exact code has not been tested but a pretty close analogue has been against my personal Github from my PC. I expect to find some little problems when I actually install it in production on the hmftools repository.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -2,7 +2,7 @@ steps:
   - id: 'Fetch token for github'
     name: gcr.io/cloud-builders/gcloud
     entrypoint: 'bash'
-    args: [ '-c', "gcloud secrets versions access latest --secret=hmftools-github-release-token --format='get(payload.data)' | tr '_-' '/+' | base64 -d > /workspace/github.token" ]
+    args: [ '-c', "gcloud secrets versions access latest --secret=hmftools-github-private-key --format='get(payload.data)' | tr '_-' '/+' | base64 -d > /workspace/github.key" ]
   - id: 'Fetch password for Dockerhub'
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -2,15 +2,15 @@ steps:
   - id: 'Fetch token for github'
     name: gcr.io/cloud-builders/gcloud
     entrypoint: 'bash'
-    args: [ '-c', "gcloud secrets versions access latest --secret=hmftools-github-release-token --format='get(payload.data)' | tr '_-' '/+' | base64 -d > /workspace/github.token" ]
+    args: [ '-c', "gcloud secrets versions access latest --secret=hmftools-github-private-key --format='get(payload.data)' | tr '_-' '/+' | base64 -d > /workspace/github.key" ]
```
mvanniekerkHartwig, line 5: Maybe add the correct gcp project for the secret here?

mvanniekerkHartwig (approved):

PR #607, opened 2024-09-02
ACTIN-1142: Ensure only exonic variants that are phased with reported variants are shown in ORANGE

cbruel:
A few weeks ago I noticed (due to an ORANGE report I was reviewing before sharing) that we were stating in the README in the “Other potentially relevant variants” section: _"**Exonic** variants that are not reported but are phased with variants that are reported"_, but in code we were actually not checking for exonic/intronic. Therefore I removed "Exonic" from the README to be in line with our code, but I do think it would be better to only show exonic variants on the report, so I added this logic now

### DIFF
--- a/orange/README.md
+++ b/orange/README.md
@@ -135,7 +135,7 @@ interesting and added to the report:
 
 - Other potentially relevant variants:
     1. Variants that are hotspots or near hotspots but not part of the reporting gene panel.
-    2. Variants that are not reported but are phased with variants that are reported.
+    2. Exonic variants that are not reported but are phased with variants that are reported.
     3. Variants that are considered relevant for tumor type classification according to Cuppa.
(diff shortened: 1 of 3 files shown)


kzuberihmf (approved):
LGTM. Not sure if a non-null exon can ever be zero, but the test seems safe anyway.

kduyvesteyn (approved):
Looks good to me as well!

Wrt Khalid's feedback: exon can never be < 1 so I think I'd remove this check.

You could consider throwing an exception in case exon rank < 1 but that should ideally happen when we produce the object, not when we use it. 

Another consideration: Any variant with an affected exon > 0 will have a gene, so you could consider removing that check, though don't have a strong opinion - the gene check does make the code more readable and resilient I think.

PR #612, opened 2024-09-14
ACTIN-292: Add Chromosomal Rearrangements

cbruel:


orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
```diff
@@ -0,0 +1,93 @@
...
+import com.hartwig.hmftools.common.genome.refgenome.RefGenomeCoordinates;
+import com.hartwig.hmftools.datamodel.orange.OrangeRefGenomeVersion;
+
+import org.jetbrains.annotations.NotNull;
+
+final class ChromosomalRearrangementsDeterminer
+{
+    private final RefGenomeCoordinates refGenomeCoordinates;
+
+    public ChromosomalRearrangementsDeterminer(@NotNull final OrangeRefGenomeVersion refGenomeVersion)
```
kduyvesteyn, line 15: Minor but we typically only use constructors that take the exact properties as input.

This function I would make into a public static factory method (e.g. see `RefGenomeVersion.from`)

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public static final double LOWER_THRESHOLD_TRISOMY = 2.8;
+    public static final double UPPER_THRESHOLD_TRISOMY = 3.5;
+    public static final double THRESHOLD_DELETION = 0.2;
+    public static final double PCT_98 = 0.98;
+    public static final double PCT_90 = 0.90;
+
+    private static final String CHR_1 = "1";
+    private static final String CHR_19 = "19";
```
kduyvesteyn, line 27: This logic won't work if the pipeline is run in v38 mode since the chromosomes will be prefixed with `chr` in that case. There are multiple ways to handle this, one way would be to always strip the prefix prior to comparison with your constants (see `RefGenomeFunctions.stripChrPrefix`)

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleInterpreter.java
```diff
@@ -68,21 +71,25 @@ public class PurpleInterpreter
     private final List<DriverGene> driverGenes;
     @NotNull
     private final LinxRecord linx;
+    @NotNull
+    private final OrangeRefGenomeVersion refGenomeVersion;
```
kduyvesteyn, line 75: I would suggest to create the `ChromosomalRearrangementsDeterminer` prior to this class and passing that in as a parameter. That way you don't have to pass around more basic properties such as the ref genome version. Those kind of complexities should be captured by the logic inside the classes that are passed in.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleInterpreter.java
```diff
@@ -444,4 +452,15 @@ private static PurpleCharacteristics createCharacteristics(@NotNull PurpleData p
                 .svTumorMutationalBurden(purple.purityContext().svTumorMutationalBurden())
                 .build();
     }
+
+    @NotNull
+    private static ChromosomalRearrangements createSuspectChromosomalRearrangments(@NotNull PurpleData purple,
```
kduyvesteyn, line 457: Why are these arrangements "suspect"? 

It may actually be suspect but that doesn't reflect anywhere else in the code. I would either go for suspect-all-the-way or not-at-all.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminerTest.java
```diff
@@ -0,0 +1,103 @@
...
+import com.hartwig.hmftools.common.purple.PurpleCopyNumber;
+import com.hartwig.hmftools.common.purple.PurpleTestUtils;
+import com.hartwig.hmftools.datamodel.orange.OrangeRefGenomeVersion;
+
+import org.jetbrains.annotations.NotNull;
+import org.junit.Test;
+
+public class ChromosomalRearrangementsDeterminerTest
+{
+    private static final ChromosomalRearrangementsDeterminer chromosomalRearrangementsDeterminer =
```
kduyvesteyn, line 20: private static final variables should be named like `CHROMOSOMAL_REARRANGEMENTS_DETERMINER`

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void canDetermine1qTrisomy()
+    {
+        int centromere1 = refCoordinates.centromere("1");
+        int endChromosome1 = refCoordinates.length("1");
+
+        int stepSize = (endChromosome1 - centromere1) / 4;
+        List<Integer> steps = createSteps(stepSize, centromere1 - 3000000, endChromosome1);
+
+        final List<PurpleCopyNumber> arm1Q = Lists.newArrayList(
```
kduyvesteyn, line 33: We do not use final keyword for local variables in orange

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminerTest.java
kduyvesteyn, line 1: Ideally you would want simpler tests (e.g. with reference data where chromosome 1 is 10 bases long with centromere in the middle), but that's not trivial (you can't depend on RefGenomeCoordinates anymore). I think I would leave as-is but just wanted to mention that tests should ideally use simple reference data if the real reference data is harder to read.

In addition, you could decide to make one test using v38 ref genome and another test v37 to make sure it works for both cases.

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminerTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        assertTrue(chromosomalRearrangementsDeterminer.determine1p19qCodeletion(with1pDeletionAndWith19QDeletion));
+        assertFalse(chromosomalRearrangementsDeterminer.determine1p19qCodeletion(with1pDeletionAndWithout19QDeletion));
+        assertFalse(chromosomalRearrangementsDeterminer.determine1p19qCodeletion(without1pDeletionAndWith19QDeletion));
+    }
+
+    @NotNull
+    private static PurpleCopyNumber copyNumber(String chromosome, int start, int end, double copyNumber)
+    {
+        return PurpleTestUtils.createCopyNumber(chromosome, start, end, copyNumber).build();
```
kduyvesteyn, line 90: Don't think I'd have this function since the PurpleTestUtils function is already quite concise, but fairly minor and not super strong opinion

orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/PurpleRecord.java
```diff
@@ -80,4 +80,7 @@ public interface PurpleRecord
 
     @Nullable
     List<PurpleLossOfHeterozygosity> reportableGermlineLossOfHeterozygosities();
+
+    @NotNull
+    ChromosomalRearrangements chromosomalRearrangements();
```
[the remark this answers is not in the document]
cbruel: @kduyvesteyn That was actually my first try. But rethinking it, I think it does not fit 100%, so I moved it. Will move it under PurpleCharacteristics though

orange/src/test/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminerTest.java
cbruel: I do like it :)

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
```diff
@@ -0,0 +1,93 @@
...
+    public ChromosomalRearrangementsDeterminer(@NotNull final OrangeRefGenomeVersion refGenomeVersion)
+    {
+        refGenomeCoordinates = getRefGenomeCoordinates(refGenomeVersion);
+    }
+
+    public static final double LOWER_THRESHOLD_TRISOMY = 2.8;
+    public static final double UPPER_THRESHOLD_TRISOMY = 3.5;
+    public static final double THRESHOLD_DELETION = 0.2;
+    public static final double PCT_98 = 0.98;
+    public static final double PCT_90 = 0.90;
```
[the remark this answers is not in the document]
cbruel: @kduyvesteyn Yes I did consider/try, but it's difficult to come up with a correct descriptive name, since e.g. 0.98 is used by both trisomy and codeletion logic and both percentages are lower limits.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
cbruel: Thanks for catching this!

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
cbruel: Could you elaborate what you mean here? Trying to understand and tried some things, but I can't get it right

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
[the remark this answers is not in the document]
kduyvesteyn: I would use 2 different variables for this actually.

The fact that they are both 0.98 is a detail, they represent different things (and you could imagine you may want to tweak one without touching the other, that would currently not be possible.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
kduyvesteyn: I have just pushed the change, please take a look! Easier to explain that way then by commenting I thought.

cbruel: Yes, thanks, this helps to understand it!

kduyvesteyn (approved):
Some new minor comments but still approved :)

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/ChromosomalRearrangementsDeterminer.java
```diff
@@ -0,0 +1,107 @@
...
+
+        return (double) lengthAboveLowerThreshold / length1QArm > LOWER_LIMIT_PERCENTAGE_TRISOMY
+                && (double) lengthBelowUpperThreshold / length1QArm > UPPER_LIMIT_PERCENTAGE_TRISOMY;
+    }
+
+    public boolean determine1p19qCodeletion(@NotNull List<PurpleCopyNumber> allSomaticCopyNumbers)
+    {
+        int centromere1 = refGenomeCoordinates.centromere(CHR_1);
+        int centromere19 = refGenomeCoordinates.centromere(CHR_19);
+        int length19QArm = refGenomeCoordinates.length(CHR_19) - centromere19;
```
cbruel, line 73: @kduyvesteyn Should there be a "+ 1" added here? Same in line 43.

cbruel: @kduyvesteyn I made some drawings and looked into how this is calculated in SvUtilities, and figured it out :)

PR #620, opened 2024-10-09
ACTIN-346:  Derive 5 breakend fields from root sources in ORANGE

kzuberihmf:
* Introduces LinxBreakendInterpreter for the derivations
* Adds structural variants and Ensembl cache (for karyotypes) as inputs to LinxInterpreter

Draft, currently falls back to stubbing out the fields in some corner cases.


orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
```diff
@@ -0,0 +1,140 @@
...
+    static double junctionCopyNumber(
+            @NotNull com.hartwig.hmftools.common.linx.LinxBreakend linxBreakend,
+            @NotNull StructuralVariant sv,
+            @NotNull LinxSvAnnotation svAnnotation
+    )
+    {
+        if(sv.type() == StructuralVariantType.SGL && linxBreakend.regionType() == com.hartwig.hmftools.common.gene.TranscriptRegionType.IG)
+        {
+            return 0D;
+        }
```
kzuberihmf, line 103: @DFKoetsier this is what I meant by SGL & IG. Was just a quick observation from running on some tests that cases with 0 had these attributes. Haven't worked out why or if comprehensive.

hmf-common/src/main/java/com/hartwig/hmftools/common/genome/chromosome/CytoBands.java
```diff
@@ -55,7 +57,7 @@ private void loadRegions(final RefGenomeVersion refGenomeVersion)
         int posEndIndex = 2;
         int nameIndex = 3;
 
-        for(String line : lines)
+        lines.stream().skip(1).forEach(line ->
         {
```
kzuberihmf, line 61: Ended up not using this, but when testing it was failing on handling it's data files header-line, hence the tiny drive-by fix. Guess it needs a unit test as well.

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
```diff
@@ -0,0 +1,140 @@
...
+        this.ensemblDataCache = ensemblDataCache;
+    }
+
+    public LinxBreakend interpret(@NotNull com.hartwig.hmftools.common.linx.LinxBreakend linxBreakend)
+    {
+        ImmutableLinxBreakend.Builder builder = breakendBuilder(linxBreakend);
+        LinxSvAnnotation svAnnotation = linxSvAnnotations.stream()
+                .filter(s -> s.svId() == linxBreakend.svId())
+                .findFirst()
+                .orElse(null);
```
kzuberihmf, line 48: Might be worth replacing the linear searches with a map lookup I suppose

DFKoetsier (commented):
A few comments, but otherwise looks good to me.

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
kzuberihmf: You suggested elsewhere that if the simple average looks reasonable perhaps we could stick with it. I'm inclined now to simplify the code and just use the average. I've found another example of the min=0 special case where it behaves differently, and I don't want to further complicate.

hmf-common/src/main/java/com/hartwig/hmftools/common/genome/chromosome/CytoBands.java
kzuberihmf: Added test

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
kzuberihmf: Now using maps

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
```diff
@@ -0,0 +1,140 @@
...
+            LOGGER.warn("No chromosome found for breakend {}, using fallbacks for derived attributes", linxBreakend.id());
+            return builder.build();
+        }
+
+        Byte orientation = sv.orientation(linxBreakend.isStart());
+        if(orientation == null)
+        {
+            LOGGER.warn("No orientation found for breakend {}, using fallbacks for derived attributes", linxBreakend.id());
+            return builder.build();
+        }
```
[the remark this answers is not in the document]
kzuberihmf: Good point. I've revised it it no longer applies blanket fallbacks to all the derived fields

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
```diff
@@ -0,0 +1,206 @@
...
+        com.hartwig.hmftools.datamodel.linx.LinxBreakend left = interpreter.interpret(breakends.get(0));
+        assertNotNull(left);
+        assertEquals("7", left.chromosome());
+        assertEquals("q34", left.chromosomeBand());
+        assertEquals(LinxBreakendType.BND, left.type());
+        assertEquals(1, left.orientation());
+        assertEquals(1.5D, left.junctionCopyNumber(), EPSILON);
+
+        com.hartwig.hmftools.datamodel.linx.LinxBreakend right = interpreter.interpret(breakends.get(1));
+        assertFallbackBreakendAttributes(right);
```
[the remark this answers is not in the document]
kzuberihmf: Removed this test, not sure what I was thinking. The breakends are handled individually here.

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
```diff
@@ -0,0 +1,161 @@
...
+    @NotNull
+    private static LinxSvAnnotation createSvAnnotation(double junctionCopyNumberMin, double junctionCopyNumberMax)
+    {
+        return LinxTestFactory.svAnnotationBuilder()
+                .svId(1)
+                .vcfId(VCF_ID)
+                .junctionCopyNumberMin(junctionCopyNumberMin)
+                .junctionCopyNumberMax(junctionCopyNumberMax)
+                .build();
+    }
```
DFKoetsier, line 153: For some of the above helper functions IDEA is warning that the variables always have the same value. You could consider varying these values a bit more between tests to show that they don't matter, or maybe just remove the option to set them in the test and hardcode them to something sensible in the helper function. I would be fine with changing nothing as well.

DFKoetsier (commented):
Two very minor comments. I would also be fine with merging as is, if you prefer that.

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
```diff
@@ -0,0 +1,161 @@
...
+        assertEquals(LinxBreakendType.BND, right.type());
+        assertEquals(0, right.orientation());
+        assertEquals(1.5D, right.junctionCopyNumber(), 0.0D);
+    }
+    
+    @Test
+    public void canComputeJunctionCopyNumber()
+    {
+        LinxBreakend linxBreakend = createBreakend(1, 1, true, TranscriptRegionType.INTRONIC);
+        StructuralVariant sv = createStructuralVariant(StructuralVariantType.DUP);
```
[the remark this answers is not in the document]
kzuberihmf: Thanks, fixed.

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
kzuberihmf: Removed constant arguments and set values in the helpers.

DFKoetsier (approved):
LGTM!

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreter.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class LinxBreakendInterpreter
+{
+
+    @NotNull
+    Map<String, StructuralVariant> structuralVariantsMap;
+    @NotNull
+    Map<Integer, LinxSvAnnotation> linxSvAnnotationsMap;
+    @NotNull
+    EnsemblDataCache ensemblDataCache;
```
kduyvesteyn, line 30: Variables should be `private final`

orange/src/main/java/com/hartwig/hmftools/orange/algo/linx/LinxInterpreter.java
```diff
@@ -3,30 +3,60 @@
...
+    @NotNull
+    List<StructuralVariant> allSomaticStructuralVariants;
+    @Nullable
+    List<StructuralVariant> allGermlineStructuralVariants;
+    @NotNull
+    List<StructuralVariant> allInferredSomaticStructuralVariants;
+    @Nullable
+    List<StructuralVariant> allInferredGermlineStructuralVariants;
+    @NotNull
+    EnsemblDataCache ensemblDataCache;
```
kduyvesteyn, line 41: Variables should be `private final`

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleData.java
```diff
@@ -44,6 +44,12 @@ public interface PurpleData
     @Nullable
     List<StructuralVariant> allGermlineStructuralVariants();
 
+    @NotNull
+    List<StructuralVariant> allInferredSomaticStructuralVariants();
```
kduyvesteyn, line 48: It is not clear from the datamodel that the inferred somatic structural variants are not included in `allSomaticStructuralVariants`, this is somewhat confusing! 

I think the clean solution would be to load both PASS and INFERRED sv's into the `all` variable and then downstream users have to filter, but might be more convenient to just rename the `allSomatic` variables to `allPassSomatic`?

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleDataLoader.java
```diff
@@ -90,6 +94,8 @@ private static PurpleData load(@NotNull String tumorSample, @Nullable String ref
         List<StructuralVariant> allSomaticStructuralVariants =
                 StructuralVariantFileLoader.fromFile(somaticStructuralVariantVcf, new PassingVariantFilter());
 
+        List<StructuralVariant> allInferredSomaticStruturalVariants = loadInferredStructuralVariants(somaticStructuralVariantVcf);
```
kduyvesteyn: Feels a bit silly that we load an entire VCF twice. Can't we load once and then apply filtering?

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
```diff
@@ -0,0 +1,161 @@
...
+                createStructuralVariant()
+        );
+
+        List<LinxSvAnnotation> linxSvAnnotations = List.of();
+        List<LinxBreakend> breakends = createBreakends();
+
+        LinxBreakendInterpreter interpreter = createInterpreter(structuralVariants, linxSvAnnotations);
+        com.hartwig.hmftools.datamodel.linx.LinxBreakend left = interpreter.interpret(breakends.get(0));
+        assertEquals("", left.chromosome());
+        assertEquals("q34", left.chromosomeBand());
```
kduyvesteyn, line 62: Haven't looked in detail but seems weird to have a chromosome band without a chromosome? Is this intended behaviour?

kduyvesteyn (approved):
Have put my faith in David's review, did notice some room for a few technical improvements. Also made a few (formatting) changes that I already committed

kduyvesteyn:
Did you test your changes on a real orange json? I imagine the JSON should be identical after this change, is that correct?  Should be testable on at least the COLO json which is part of the repo.

orange/src/main/java/com/hartwig/hmftools/orange/algo/purple/PurpleDataLoader.java
kzuberihmf: Yes this was bothering me also, but introducing inadvertent regressions while working on the basic implementation bothered me more 😄. I think worth improving, will have a look.

orange/src/test/java/com/hartwig/hmftools/orange/algo/linx/LinxBreakendInterpreterTest.java
kzuberihmf: The chromosome is extracted from the associated structural variant, switching on if the breakend is start or not. Turns out the chromosome band (following the original derivation) just comes from the gene via the ensembl cache. Since the gene is available in the breakend, it's technically possible for us to fill this in without the chrom if the structural variant can not be resolved for some reason (will result in warning in the log). It does look weird.

kduyvesteyn: Clear. Don't think its super important to "fix", hopefully the situation is rare anyways :)

kzuberihmf:
Testing a backport of this update (to orange 3.7.0) on full COLO from pipeline 5, the only differences are to junction copy number:

* small 10E-4 or smaller differences, due to this change not rounding, and the previous version loading a rounded double from the breakend tsv. This difference seems trivial, can round if desired.
* 4 breakends (related to the same sv) having a junction copy number of 0.0185 now but 0.037 before. This is due to the edge case that @DFKoetsier described where min jcn is 0.0 and the max was used instead of the average. This change always computes the simple average instead of attempting to reproduce the edge cases (there are others) from the original calculation

Hope that makes sense.

kduyvesteyn:
> Testing a backport of this update (to orange 3.7.0) on full COLO from pipeline 5, the only differences are to junction copy number:
> 
> * small 10E-4 or smaller differences, due to this change not rounding, and the previous version loading a rounded double from the breakend tsv. This difference seems trivial, can round if desired.
> * 4 breakends (related to the same sv) having a junction copy number of 0.0185 now but 0.037 before. This is due to the edge case that @DFKoetsier described where min jcn is 0.0 and the max was used instead of the average. This change always computes the simple average instead of attempting to reproduce the edge cases (there are others) from the original calculation
> 
> Hope that makes sense.

Will go with David's judgment here. All your other changes look good to me! 

Made a number of formatting changes that weren't picked up by auto-format, not sure why.

PR #626, opened 2024-10-30
INFRA-515

kzagranovskyi-hw:
Updated tools data loading to be independent of file naming


kzagranovskyi-hw:
Updated tools data loading to be backward compatible

mvanniekerkHartwig (approved):

PR #627, opened 2024-10-30
INFRA-515 Updated tools data loading to be independent of file naming

kzagranovskyi-hw:


kzagranovskyi-hw:
Updated tools data loading to be backward compatible

mvanniekerkHartwig (changes requested):
hmftools is using a different formatting convention from the rest of our code. Better to stick with what they use.

PR #634, opened 2024-11-12
Esvee: DEV-4113: Make sample argument optional in Caller config again

DFKoetsier:
Needed to make germline-only runs work. Germline-only runs used to work, but then this argument was set back to required.


DFKoetsier:
Please feel free to ask someone else to do this review instead. I just didn't know who I should choose.

PR #650, opened 2024-12-09
ACTIN-1551: Make gene copy number datamodel in ORANGE transcript-aware

cbruel:


orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/purple/PurpleGeneCopyNumber.java
```diff
@@ -13,6 +13,11 @@ public interface PurpleGeneCopyNumber
     @NotNull
     String gene();
 
+    @NotNull
+    String transcript();
+
+    boolean isCanonical();
+
```
kduyvesteyn, line 20: In `LinxBreakend` these fields appear after chromosome. Would suggest to make it consistent

kduyvesteyn (approved):

PR #679, opened 2025-03-10
ACTIN-1779: Return null instead of emptyList in JsonFunctions optiona…

spdeleeuw:
…lStringList


kduyvesteyn (approved):
Looks good to me!

PR #687, opened 2025-04-02
ACTIN-1845: Update test data and example reports in ORANGE to pipeline v6

kduyvesteyn:
Structural changes:
 - Renamed PEACH file from “genotypes” to “haplotypes.best”
 - Dropped previous flagstat files and point flagstats to new metrics dir

@charlesshale @DFKoetsier I compared against actual pipeline call and data and couldn't find other more structural cleanups but if I missed anything let me know!

@zoehehartwig no need for review (assume @DFKoetsier will review) but would be good to be aware of this setup


orange/src/test/resources/test_run/chord/tumor_sample.chord.prediction.tsv
```diff
@@ -1,2 +1,2 @@
 sample	p_BRCA1	p_BRCA2	p_hrd	hr_status	hrd_type	remarks_hr_status	remarks_hrd_type
-tumor_sample	0	0	0	HR_proficient	none		
+tumor_sample	0	0	0	HR_proficient	none
```
DFKoetsier, line 2: Strictly speaking the tabs should be there, since they are there in the actual pipeline output file as well because the last two columns are empty.

DFKoetsier (commented):
Two small things. Otherwise looks good to me.

orange/src/test/resources/test_run/chord/tumor_sample.chord.prediction.tsv
kduyvesteyn: Thanks, good catch, have added them back in

orange-datamodel/src/test/resources/real.orange.json
```diff
@@ -397,7 +397,7 @@
         "worstCodingEffect": "MISSENSE",
         "canonicalImpact": {
           "transcript": "ENST00000355716",
-          "hgvsCodingImpact": "c.50A>G",
+          "hgvsCodingImpact": "c.50A\u003eG",
```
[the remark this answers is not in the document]
kduyvesteyn: I just copy-pasted the JSON and not sure it matters but have changed it back nonetheless. If anything, its nice if this JSON is properly readable by a human :)

PR #690, opened 2025-04-08
ACTIN-1857: Replace "platinum" with "pipeline"

zoehehartwig:
I used the search and replace tool over the entire project. I am going to review it to see if there are any that shouldn't have been replaced and change them back.


kduyvesteyn (approved):
Looks good @zoehehartwig ! 

Just 2 minor comments.

Also, don't forget to include this change in the release notes. It has no functional impact but any (breaking) datamodel change is good to document for any other person depending on ORANGE (not necessarily just ACTIN)

PR #691, opened 2025-04-14
Use multiplatform UMCCR gatk-bwamem-jni instead of org.broadinstitute.hellbender.*

brainstorm:
This PR subsitutes Broad Institute's gatk-bwamem-jni for UMCCR's fork:

https://github.com/umccr/gatk-bwamem-jni/

Building native objects for 4 os-arch pairs, signing and publishing to Maven Central (Sonatype) now happens transparently in GitHub's CI/CD for greater convenience and more accessible security auditing (as opposed to non-CI/CD built artifacts), i.e:

https://github.com/umccr/gatk-bwamem-jni/actions/runs/13742549372

The original motivation for this change is seamless multiplatform support of that dependency [when working on an Linux-Aarch64 port of Oncoanalyser](https://github.com/nf-core/oncoanalyser/pull/134).

Merging this code would avoid having to use an EVSEE fork that supports Linux/Aarch64 and/or setting up `-DLIBBWA_PATH` and [shipping binary blobs](https://github.com/hartwigmedical/hmftools/blob/master/libbwwwa.Darwin.dylib) in Hmftools' code tree.

/cc @scwatts @ohofmann


ohofmann:
+1 from me here. This is likely the last blocker before we can confirm arm64 support for the WiGiTS framework which should save folks a bit of money for their large cohort re-analysis projects.

ohofmann (approved):

ohofmann:
Argh. Disregard the approval, that was not intended.

PR #697, opened 2025-04-18
Orange: DEV-4172: Fix variants with only non-canonical impact

DFKoetsier:
Specifically, when there is also another variant reportable on the canonical transcript of the same gene.

### DIFF
--- a/orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
+++ b/orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
@@ -48,7 +48,7 @@ public class VariantEntryFactoryTest
 
         PurpleVariant nonDriverVariant = TestPurpleVariantFactory.builder()
                 .gene("gene 2")
-                .canonicalImpact(TestPurpleVariantFactory.impactBuilder().hgvsProteinImpact("impact 3").build())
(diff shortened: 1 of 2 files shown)


orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
```diff
@@ -182,6 +182,72 @@ public void shouldGenerateMultipleVariantsForNonCanonicalAndCanonicalDriver()
...
+
+        assertEquals(3, entries.size());
+        VariantEntry entry2 = findByGeneAndImpact(entries, "gene 1", "impact 2");
+        assertFalse(entry2.isCanonical());
+
+        VariantEntry entry3 = findByGeneAndImpact(entries, "gene 1", "impact 3");
+        assertTrue(entry3.isCanonical());
+
+        VariantEntry entry4 = findByGeneAndImpact(entries, "gene 1", "impact 4");
+        assertFalse(entry4.isCanonical());
```
kzuberihmf, line 248: very trivial: maybe could rename the entries to 1 2 & 3 instead of 2 3 & 4 (copied from previous test I think).

kzuberihmf (approved):

orange/src/test/java/com/hartwig/hmftools/orange/report/datamodel/VariantEntryFactoryTest.java
DFKoetsier: Good point, thanks!

PR #699, opened 2025-04-21
ACTIN-1898: Remove UGT1A1 from PEACH output ingested in ORANGE

zoehehartwig:
> Recently some issues have been discovered wrt PEACH calling in UGT1A1 and proposal is to not ingest it anymore in ORANGE (and remove all subsequent usage in ACTIN). 
Eventually it will be added back but the expectation is that this will take ~months and in the meantime we should not share anything wrt this gene anymore.


orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -252,30 +252,62 @@ public OrangeRecord run(@NotNull OrangeConfig config) throws Exception
...
-                .virusInterpreter(virusInterpreter != null ? VirusInterpreter.interpret(virusInterpreter) : null)
-                .chord(chord != null ? OrangeConversion.convert(chord) : null)
-                .cuppa(cuppa)
-                .peach(ConversionUtil.mapToIterable(peach, OrangeConversion::convert))
-                .sigAllocations(SigsInterpreter.interpret(sigAllocations, etiologyPerSignature))
-                .cohortEvaluations(evaluateCohortPercentiles(config, purple))
-                .plots(buildPlots(config))
-                .build();
+        OrangeRecord report;
+        if (peach != null)
```
pauldwolfe, line 256: having the null check up here introduced a lot of duplication, let's try to narrow the scope.

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -252,30 +252,62 @@ public OrangeRecord run(@NotNull OrangeConfig config) throws Exception
...
+                    .purple(purple)
+                    .linx(linx)
+                    .wildTypeGenes(wildTypeGenes)
+                    .isofox(isofox)
+                    .lilac(lilac != null ? OrangeConversion.convert(lilac, hasRefSample, config.rnaConfig() != null) : null)
+                    .immuneEscape(immuneEscape)
+                    .virusInterpreter(virusInterpreter != null ? VirusInterpreter.interpret(virusInterpreter) : null)
+                    .chord(chord != null ? OrangeConversion.convert(chord) : null)
+                    .cuppa(cuppa)
+                    .peach(ConversionUtil.mapToIterable(peach.stream().filter(genotype -> !genotype.gene().equals("UGT1A1")).toList(), OrangeConversion::convert))
```
pauldwolfe, line 277: I would suggest creating a method/function that takes the set of peach genotype as input and:
- checks for the null case
- does the filtering.

It's a bit more consistent with the rest of the class, which breaks out a method and variable when the conversion becomes more than one line (eg. loadCuppaData).

pauldwolfe (changes requested):

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -758,6 +761,18 @@ private static void verifyPlots(@NotNull OrangePlots orangePlots, @NotNull LinxD
         }
     }
 
+    private static List<PeachGenotype> peachFilterOutUGT1A1(List<PeachGenotype> peach)
```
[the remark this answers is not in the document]
cbruel: We could also include the filtering logic in the `loadPeachData` function? So we don't need an extra function. But okay with keeping it like this, if you both think this is better

orange/src/main/java/com/hartwig/hmftools/orange/algo/OrangeAlgo.java
```diff
@@ -758,6 +759,18 @@ private static void verifyPlots(@NotNull OrangePlots orangePlots, @NotNull LinxD
...
```
cbruel, line 762: I think it's more in line with other function names to call this something like `"filterUGT1A1FromPeach"`

cbruel (commented):
- Could you add your change to the README? (i.e. adding your change to the 'Upcoming' section. Also still needed for your previous ORANGE change (adding the cuppa mode))
- When the report changes (in this case the UGT1A1 is removed from the report), we have to update the COLO829 test reports in `orange > src > main > resources`. You can do this by running `ReportGeneratorTestApplication` (To create the WGS report: Tumor_only=false, To create the targeted report: Tumor_only=true)
- You can also use that script to update `real.orange.json` (see comment in script). That way the “samplingDate” is adjusted automatically to the current date

PR #727, opened 2025-07-14
GeneUtils: PanelBuilder: redesign and feature expansion

reecejones-hartwig:


hmf-common/src/main/java/com/hartwig/hmftools/common/utils/file/DelimFileReader.java
```diff
@@ -219,7 +219,20 @@ public Row next()
                     String line = nextLine;
                     nextLine = null;
                     assert line != null;
-                    String[] values = line.split(mDelim, -1);
+
```
reecejones-hartwig, line 222: This algorithm doesn't support regex delimiter like the previous algorithm did. However, I don't think `DelimFileReader` intends to support a regex delimiter. There are no uses of it with a regex within the codebase. So it should be safe to make this change. The performance increase is significant.

hmf-common/src/main/java/com/hartwig/hmftools/common/utils/file/DelimFileWriter.java
```diff
@@ -253,6 +253,13 @@ public void set(String key, String value)
         // record.set("rate", 0.27572, new DecimalFormat("#.####", new DecimalFormatSymbols(Locale.ENGLISH)));
         public void set(String key, double value, NumberFormat format) { set(key, format.format(value)); }
 
+        public void setOrNull(String key, String value) { set(key, value == null ? sNullIndicator : value); }
```
reecejones-hartwig, line 256: The existing `set()` methods assume the argument is non-null and throw on null. However the class does support writing nulls to file. So I added these methods which allow null and produce a null value in the file.

hmf-common/src/main/java/com/hartwig/hmftools/common/region/ChrBaseRegion.java
```diff
@@ -50,6 +50,17 @@ public ChrBaseRegion(final String chromosome, final int posStart, final int posE
     }
 
     public static ChrBaseRegion from(final GenomeRegion region) { return new ChrBaseRegion(region.chromosome(), region.start(), region.end()); }
+
+    public static ChrBaseRegion from(final String chromosome, final BaseRegion region)
```
reecejones-hartwig, line 54: These seem like common construction methods

PR #765, opened 2025-11-10
Patient-db: fix amber concurrent loading race condition

KasperWolsink99:
This PR aims to fix a race condition bug when loading Amber samples concurrently using PatientDb.
The bug was caused by the way how the next amber patient id is generated. To generate the next patient id, we count all existing patients and add 1 to find the next id, which is vulnerable to race conditions when you load samples currently. A simple fix is to lock all Amber-related tables during loading, which is what this PR did.

### DIFF
--- a/patient-db/src/main/java/com/hartwig/hmftools/patientdb/amber/LoadAmberData.java
+++ b/patient-db/src/main/java/com/hartwig/hmftools/patientdb/amber/LoadAmberData.java
@@ -133,7 +133,6 @@ public class LoadAmberData
         {
             LOGGER.info("releasing Amber table locks for sample {}", sample.sampleId());
             dbAccess.context().execute("UNLOCK TABLES");
-            LOGGER.info("Amber table locks released");
         }
     }
 


patient-db/src/main/java/com/hartwig/hmftools/patientdb/amber/LoadAmberData.java
```diff
@@ -97,32 +97,41 @@ public static void main(@NotNull String[] args) throws ParseException, IOExcepti
...
     {
-        LOGGER.info("comparing with existing samples");
-
-        final List<AmberSample> allSamples = dbAccess.readAmberSamples();
-        final List<AmberMapping> sampleMappings = new ArrayList<>();
-        for(AmberSample other : allSamples)
+        try
         {
-            if(!other.sampleId().equals(sample.sampleId()))
+            dbAccess.context().execute("LOCK TABLES amberPatient WRITE, amberSample WRITE, amberMapping WRITE");
```
mdenburger: Maybe add some logging before locking and unlocking tables? That way it will be more clear why a concurrently run Isofox loader "hangs" for a while.

mdenburger (approved):
Nice fix! As discussed, the addition of testcontainers may be too much, so let's wait for AUS to chime in there.

One more thing: this PR has been based on `master`, but I think we also need it in the `pipeline_v6_0` branch.

PR #772, opened 2025-12-03
Common: DelimFileReader: deduplicate logic and improve interface

reecejones-hartwig:


hmf-common/src/main/java/com/hartwig/hmftools/common/utils/file/DelimFileReader.java
```diff
@@ -293,272 +294,143 @@ private Row(final Map<String, Integer> columnIndexMap, final String[] values)
...
+            // TODO? should forbid more than 1 char
+            return getValue(column, string -> string.charAt(0), allowNull);
         }
 
-        public @Nullable Boolean getBooleanOrNull(final Enum<?> column)
+        private Boolean getBoolean(int column, boolean allowNull)
         {
-            return getBooleanOrNull(column.name());
+            // TODO: should accept only true or false
+            return getValue(column, Boolean::parseBoolean, allowNull);
```
hongwingl: yes we should make sure this gets fixed. Possibly by making our own `parseBoolean` function here.

hmf-common/src/main/java/com/hartwig/hmftools/common/utils/file/DelimFileReader.java
```diff
@@ -293,272 +294,143 @@ private Row(final Map<String, Integer> columnIndexMap, final String[] values)
...
+            else
             {
-                return null;
+                return rawValue;
             }
-            return v;
         }
 
-        public int getInt(final int columnIndex)
+        private int getColumn(Enum<?> column)
```
hongwingl: `getColumnIndex` would be a more descriptive function name.

PR #776, opened 2026-01-16
Cider: use read mates in assembly

reecejones-hartwig:
- Allow mates of anchor matched reads to extend the VDJ assembly sequence
- Delete unused code

### DIFF
--- a/cider/src/main/java/com/hartwig/hmftools/cider/AlignmentUtil.kt
+++ b/cider/src/main/java/com/hartwig/hmftools/cider/AlignmentUtil.kt
@@ -85,9 +85,8 @@ object AlignmentUtil
             .withGapExtend(-GAP_EXTEND_SCORE)
             .withExpectedValueCutoff(expectedValueCutoff)
             .withNumThreads(numThreads)
-            .withSubjectBestHit(false)
+            .withSubjectBestHit(true)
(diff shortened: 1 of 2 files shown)


cider/src/main/java/com/hartwig/hmftools/cider/VDJSequenceBuilder.kt
```diff
@@ -427,6 +437,10 @@ class VDJSequenceBuilder(private val vjLayoutAdaptor: IVJReadLayoutAdaptor,
         val combinedVjLayout: ReadLayout = ReadLayout.merge(vLayout, jLayout, 0, jAlignedPositionShift, minBaseQuality)
         combinedVjLayout.id = "${vLayout.id};${jLayout.id}"
 
+        // Extend the layout sequence by adding mates of reads in the layout.
+        // These mates wouldn't have been included earlier because they are not close enough to the anchor.
+        tryExtendCombinedLayoutWithMates(combinedVjLayout, mateLayoutReads, vLayoutGeneType, jLayoutGeneType, jAlignedPositionShift)
```
reecejones-hartwig: Because the mates are added here, after merging the V and J layouts, they won't show up in the layout output file. They also won't contribute to a few stats e.g. "V aligned read count".
Is this a problem?

PR #795, opened 2026-05-06
Opr 213

LiekeSchoenmaker:
SCOPE: 
- set low driver virus from candidate to reportable (low driver virus will be reported in oncoact) 
- set for gene disruptions/fusions/variants CN < 0.1 (rounded) to candidate. Add a warning to investigate when this occurs in prduction


finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,135 @@
...
+    @NotNull
+    public static List<SmallVariant> convertSmallVariant(@NotNull List<SmallVariant> smallVariants) {
+        List<SmallVariant> smallVariantList = new ArrayList<>();
+        for (SmallVariant smallVariant : smallVariants) {
+            Double copyNumber = roundCopyNumber(smallVariant.adjustedCopyNumber());
+
+            if (copyNumber != null && copyNumber >= 0.1) {
+                smallVariantList.add(SmallVariantBuilder.builder(smallVariant)
+                        .driver(DriverFieldsBuilder.builder(smallVariant.driver()).reportedStatus(ReportedStatus.REPORTED).build())
+                        .build());
```
ArnoHMF: I think that in this case the variant should be passed on unchanged. This applies to all similar cases in this class.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                smallVariantList.add(SmallVariantBuilder.builder(smallVariant)
+                        .driver(DriverFieldsBuilder.builder(smallVariant.driver()).reportedStatus(ReportedStatus.CANDIDATE).build())
+                        .build());
+                LOGGER.warn("Gene {} with variant (coding) {} with variant (protein) {} has lesser than <0.1 copies, investigate further",
```
ArnoHMF: Logging text should be less than instead of lesser than. This applies to all similar cases in this class.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,135 @@
...
+
+    //LS: Duplicate of the onco master repo. Is this necessary or can we compare to 0.05?
+    @Nullable
+    public static Double roundCopyNumber(@Nullable Double value) {
+        if (value != null) {
+            double doubleValue = Math.max(value, 0);
+            return Doubles.round(doubleValue, doubleValue > 10 ? 0 : 1);
+        }
+        return null;
+    }
```
ArnoHMF: This is a good question how to go about it. What you are saying here that a value of 0.05 should still be included as it gets rounded to 0.1 and you want to filter out values below 0.1. 

Apart from the question what makes most sense (I guess 0 copy numbers do not make sense so that's why we have the filtering), we want to make sure that things are consistent.

What I would probably do is move the actual rounding to this level as well in a CopyNumberRoundingConverter, and then run this converter after that. Then you can use 0.1 here and then it would be best to swap the equation to < 0.1 instead of >= 0.1

finding-datamodel/pom.xml
```diff
@@ -46,10 +46,15 @@
             <artifactId>annotations</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>${apache.log4j.version}</version>
```
ArnoHMF: This version should not be necessary because already defined in root pom

finding-datamodel/pom.xml
```diff
@@ -46,10 +46,15 @@
...
+        </dependency>
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-slf4j-impl</artifactId>
-            <scope>test</scope>
+            <version>${apache.log4j.version}</version>
```
ArnoHMF: This version should not be necessary because already defined in root pom

finding-datamodel/pom.xml
```diff
@@ -46,10 +46,15 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>${apache.log4j.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-slf4j-impl</artifactId>
-            <scope>test</scope>
```
ArnoHMF, line 52: This should be here, as this is only necessary in the tests this implementation does not need to be transitive.

finding/src/main/java/com/hartwig/hmftools/finding/ConversionUtil.java
```diff
@@ -20,23 +20,21 @@
...
-                        NoGermlineConverter::convert,
-                        CandidateToReportableConverter::convert,
-                        ReportedOnlyConverter::convert))
-                .apply(findingRecord);
+                LowPurityConverter::convert,
+                PTOConverter::convert,
+                GainDeletionsFilterConverter::convert,
+                NoGermlineConverter::convert,
+                CandidateToReportableConverter::convert,
+                CopyNumberConverter::convert,
```
ArnoHMF, line 37: This means it will also be included in the version in master. I think that is fine? Otherwise we need to make this optional somehow.

ArnoHMF (changes requested):

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
LiekeSchoenmaker: indeed, have updated

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
LiekeSchoenmaker: have update to add the whole object to the list

finding-datamodel/pom.xml
LiekeSchoenmaker: have changed the test to the ${apache.log4j.version}, this solves the issue otherwise the package cannot be found in the class

finding-datamodel/pom.xml
LiekeSchoenmaker: if I add the test, the package cannot be founded into CopyNumberConverter

finding/src/main/java/com/hartwig/hmftools/finding/ConversionUtil.java
LiekeSchoenmaker: what do you mean with this exactly?

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
LiekeSchoenmaker: So you mean to remove the GeneUtil.roundCopyNumber in OncoAct repo and do this converter in this place? Is that correct? In that case we should convert of all events the CN to this place

ArnoHMF: Yes, I mean to move to rounding of copy numbers from OncoAct to a converter as well and to run that converter before this one and then to not do any rounding here anymore and use the rounded cutoff < 0.1

LiekeSchoenmaker: I, have moved the rounding to the findings json as well. As, the reporter runs also with a json as input, the rounding of CN is also still needed for the oncoAct repo (so I haven't removed it from this repo)

ArnoHMF: I'm not sure if I understand what you mean here.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberRoundingConverter.java
```diff
@@ -0,0 +1,145 @@
...
+        return Doubles.round(doubleValue, doubleValue > COPY_NUMBER_ROUNDING_THRESHOLD ? 0 : 1);
+    }
+
+    @Nullable
+    public static Double roundCopyNumberNullable(@Nullable Double value)
+    {
+        if(value != null)
+        {
+            double doubleValue = Math.max(value, 0);
+            return Doubles.round(doubleValue, doubleValue > COPY_NUMBER_ROUNDING_THRESHOLD ? 0 : 1);
```
ArnoHMF: This is duplicated from above. Should call roundCopyNumber. And then you move this function above the roundCopyNumber

finding-datamodel/pom.xml
ArnoHMF: You need to add

        <dependency>
            <groupId>org.apache.logging.log4j</groupId>
            <artifactId>log4j-core</artifactId>
        </dependency>

That is the interface to use for the logger

finding-datamodel/pom.xml
ArnoHMF: If you add the interface module as in previous comment then this should work with scope test and without the version number I think

finding/src/main/java/com/hartwig/hmftools/finding/ConversionUtil.java
ArnoHMF: This means that the current master, once these changes are merged into hmftools, will also use this. 

And the current master is what we will put into production first.

Your other changes are in the ivdr branch.

But I think this is all fine

ArnoHMF (changes requested):

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
ArnoHMF: Never mind I saw your changes and that looks good

ArnoHMF: However I do believe the rounding can be removed from OncoAct. The assumption is now that the findings.json used is produced using this converter. We need to think a bit about what is the best setup. The alternative is that we run these converters from within OncoAct itself.

finding/src/main/java/com/hartwig/hmftools/finding/ConversionUtil.java
hongwingl: I merged those changes into master. Sorry was a little late for this!

ArnoHMF: Sorry for the confusion here but I meant the OncoAct master. I was not very clear here. 

Looking at this now I think it's better to move all the code in this package into OncoAct

finding-datamodel/pom.xml
LiekeSchoenmaker: now it works indeed

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberRoundingConverter.java
LiekeSchoenmaker: have re-used the code of the other function

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
LiekeSchoenmaker: We cannot removed the rounding currently, because the tests are now failing because orange is used as input and the output is not rounded then. We can solve this indeed to run the converters inside oncoact

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @NotNull
+    public static List<Disruption> filterDisruption(@NotNull List<Disruption> disruptions)
+    {
+        List<Disruption> disruptionList = new ArrayList<>();
+
+        for(Disruption disruption : disruptions)
+        {
+            if(disruption.disruptedCopyNumber() >= 0.1)
```
ArnoHMF: This should be a constant

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+            {
+                disruptionList.add(disruption);
+            }
+            else
+            {
+                disruptionList.add(DisruptionBuilder.builder(disruption)
+                        .driver(DriverFieldsBuilder.builder(disruption.driver()).reportedStatus(ReportedStatus.CANDIDATE).build())
+                        .build());
+                LOGGER.warn("Gene disruption {} with type {} has less than <0.1 copies, investigate further",
+                        disruption.gene(),
```
ArnoHMF, line 114: Reuse contant for 0.1 (and put a space in between < 0.1)

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @NotNull
+    public static List<Fusion> filterFusion(@NotNull List<Fusion> fusions)
+    {
+        List<Fusion> fusionsList = new ArrayList<>();
+
+        for(Fusion fusion : fusions)
+        {
+            String fusionName = fusion.geneUp() + " - " + fusion.geneDown();
+            if(fusion.junctionCopyNumber() >= 0.1)
```
ArnoHMF: Reuse constant

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+            {
+                fusionsList.add(fusion);
+            }
+            else
+            {
+                fusionsList.add(FusionBuilder.builder(fusion)
+                        .driver(DriverFieldsBuilder.builder(fusion.driver()).reportedStatus(ReportedStatus.CANDIDATE).build())
+                        .build());
+                LOGGER.warn("Fusion {} has less than <0.1 copies, investigate further", fusionName);
```
ArnoHMF: Reuse constant

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+        for(Disruption disruption : disruptions)
+        {
+            if(disruption.disruptedCopyNumber() >= 0.1)
+            {
+                disruptionList.add(disruption);
+            }
+            else
+            {
+                disruptionList.add(DisruptionBuilder.builder(disruption)
+                        .driver(DriverFieldsBuilder.builder(disruption.driver()).reportedStatus(ReportedStatus.CANDIDATE).build())
```
ArnoHMF: You can put this into a function (toCandidate()) and reuse that below

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
```diff
@@ -0,0 +1,196 @@
...
+        for(HlaAllele hlaAllele : hlaAlleles)
+        {
+            hlaList.add(HlaAlleleBuilder.builder(hlaAllele).tumorCopyNumber(roundCopyNumberNullable(hlaAllele.tumorCopyNumber()))
+                    .build());
+        }
+        return hlaList;
+    }
+
+    @NotNull
+    public static Double roundCopyNumber(@NotNull Double value)
```
ArnoHMF: Move this below the nullable version as it's used by the nullable version

ArnoHMF (changes requested):
Few comments. If you fix these then we can merge

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/CopyNumberConverter.java
LiekeSchoenmaker: good point, I will create it

PR #796, opened 2026-05-07
REDUX: RNA Transcript Fusion & Lift Back

shiv-hartwig:
**SpliceFastaBuilder** 

Output: `splice.transcript_contigs.fasta+ splice.transcript_contigs.sidecar.tsv (contig → gene, transcript, chromosome, exon spans)`

**SpliceLiftBack** 

per read; reads the sidecar tsc, translates transcript-coordinate alignments in input BAM back to genome coordinates (CIGAR N gaps for introns)

### DIFF
--- a/redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
+++ b/redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
@@ -1,0 +1,37 @@
+package com.hartwig.hmftools.redux.splice;
+
+import java.util.List;
+
+import com.hartwig.hmftools.common.region.BaseRegion;
+
+public class ContigEntry
+{
+    public final String ContigName;
+    public final String GeneId;
+    public final String GeneName;
+    public final String TransName;
+    public final String Chromosome;
+
+    // exon spans on the forward genomic strand, sorted ascending. Each span is 1-based, inclusive.
+    public final List<BaseRegion> ExonSpans;
+
+    public ContigEntry(
+            final String contigName, final String geneId, final String geneName, final String transName,
+            final String chromosome, final List<BaseRegion> exonSpans)
+    {
+        ContigName = contigName;
+        GeneId = geneId;
+        GeneName = geneName;
+        TransName = transName;
+        Chromosome = chromosome;
+        ExonSpans = exonSpans;
+    }
+
+    public int contigLength()
+    {
+        int length = 0;
+        for(BaseRegion span : ExonSpans)
+            length += span.baseLength();
+        return length;
+    }
+}
--- a/redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
+++ b/redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
@@ -4,34 +4,17 @@ import java.util.List;
 
 import com.hartwig.hmftools.common.region.BaseRegion;
 
-public class ContigEntry
+public record ContigEntry(
+        String contigName,
+        String geneId,
+        String geneName,
+        String transName,
+        String chromosome,
+        // exon spans on the forward genomic strand, sorted ascending. Each span is 1-based, inclusive.
+        List<BaseRegion> exonSpans)
 {
-    public final String ContigName;
-    public final String GeneId;
-    public final String GeneName;
-    public final String TransName;
-    public final String Chromosome;
-
-    // exon spans on the forward genomic strand, sorted ascending. Each span is 1-based, inclusive.
-    public final List<BaseRegion> ExonSpans;
-
-    public ContigEntry(
-            final String contigName, final String geneId, final String geneName, final String transName,
-            final String chromosome, final List<BaseRegion> exonSpans)
-    {
-        ContigName = contigName;
-        GeneId = geneId;
-        GeneName = geneName;
-        TransName = transName;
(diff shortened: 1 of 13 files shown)


redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
```diff
@@ -0,0 +1,37 @@
+package com.hartwig.hmftools.redux.splice;
+
+import java.util.List;
+
+import com.hartwig.hmftools.common.region.BaseRegion;
+
+public class ContigEntry
```
reecejones-hartwig: I'd suggest using a `record` for struct-like classes. It's less typing.

redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigEntry.java
```diff
@@ -0,0 +1,37 @@
...
+        GeneId = geneId;
+        GeneName = geneName;
+        TransName = transName;
+        Chromosome = chromosome;
+        ExonSpans = exonSpans;
+    }
+
+    public int contigLength()
+    {
+        int length = 0;
```
reecejones-hartwig: Stream API one-liner perhaps?

redux/src/main/java/com/hartwig/hmftools/redux/splice/ContigSidecar.java
```diff
@@ -0,0 +1,81 @@
...
+public final class ContigSidecar
+{
+    public static final String COL_CONTIG_NAME = "ContigName";
+    public static final String COL_GENE_ID = "GeneId";
+    public static final String COL_GENE_NAME = "GeneName";
+    public static final String COL_TRANS_NAME = "TransName";
+    public static final String COL_CHROMOSOME = "Chromosome";
+    public static final String COL_EXON_SPANS = "ExonSpans";
+
+    private static final List<String> COLUMNS = List.of(
```
reecejones-hartwig: `DelimFileWriter` also supports columns specified as an `Enum`, which may be less typing

redux/src/main/java/com/hartwig/hmftools/redux/splice/SpliceFastaBuilder.java
```diff
@@ -0,0 +1,156 @@
...
+        {
+            for(List<GeneData> genes : chrGeneMap.values())
+            {
+                for(GeneData gene : genes)
+                {
+                    ++genesProcessed;
+
+                    List<TranscriptData> transcripts = mEnsemblDataCache.getTranscripts(gene.GeneId);
+                    // TODO: do i really need this here? CHECK if ensembledatacache already covers this
+                    if(transcripts == null || transcripts.isEmpty())
```
reecejones-hartwig, line 71: Since the genes are coming from the Ensembl data, I'd expect this would be impossible? Should throw and fail if that's the case

redux/src/main/java/com/hartwig/hmftools/redux/splice/SpliceFastaBuilder.java
```diff
@@ -0,0 +1,156 @@
...
+                return false;
+        }
+        return true;
+    }
+
+    private static void writeFastaContig(final BufferedWriter writer, final String contigName, final String sequence) throws IOException
+    {
+        // TODO: confirm this with Charles
+        // unwrapped: matches BlastnRunner.writeBlastFasta. bwa-mem2 / samtools accept any line length.
+        writer.write(">" + contigName);
```
reecejones-hartwig, line 135: I think this format is fine. In my experience BWA does not require any specific line length. The line length is more like a guide to make FASTAs readable

redux/src/main/java/com/hartwig/hmftools/redux/splice/SpliceFastaConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class SpliceFastaConfig
+{
+    public final String EnsemblDataDir;
+    public final String RefGenomeFile;
+    public final RefGenomeVersion RefGenVersion;
+    public final String OutputDir;
+    public final String OutputId;
+
+    private boolean mIsValid;
```
reecejones-hartwig: Personally I'm not a fan of this construct object + check validity. If it's invalid it should fail immediately and not allow you to construct the config, unless there is a strong functional reason otherwise.

redux/src/main/java/com/hartwig/hmftools/redux/splice/TranscriptContigBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+        // exons are stored by Rank (transcribed order). Re-sort by genomic Start so the contig sequence is in
+        // forward-strand orientation regardless of gene strand — keeps lift-back math direct.
+        List<ExonData> ordered = new ArrayList<>(transcript.exons());
+        ordered.sort(Comparator.comparingInt(e -> e.Start));
+
+        List<BaseRegion> spans = new ArrayList<>(ordered.size());
+        StringBuilder sequence = new StringBuilder();
+        for(ExonData exon : ordered)
+        {
+            spans.add(new BaseRegion(exon.Start, exon.End));
```
reecejones-hartwig, line 47: Thoughts on a sanity check in case exons overlap or something weird like that?

PR #802, opened 2026-05-19
QC status refactoring

ArnoHMF:
Refactored status so it's split into errors and warnings to be more consistent with finding status. This also allows low purity to be an error. Replaced pass with a pass flag instead of a status entry.


hongwingl:
Looks good!

hongwingl (approved):

PR #805, opened 2026-05-29
Rna findings

hongwingl:


finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaStatistics.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Set;
+
+import com.hartwig.hmftools.finding.datamodel.finding.Finding;
+
+import jakarta.validation.constraints.NotNull;
+
+@RecordBuilder
+public record RnaStatistics(
+        @NotNull String findingKey,
+        @NotNull Set<QcStatus> qcStatus,
```
ArnoHMF: This is now inconsistent with the pattern for other statuses (warnings & errors)

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaFusion.java
```diff
@@ -0,0 +1,56 @@
...
+
+import jakarta.validation.constraints.NotNull;
+
+@RecordBuilder
+public record RnaFusion(
+        @NotNull String findingKey,
+        @Nullable String geneStart,
+        @Nullable String geneEnd,
+        @NotNull String chromosomeStart,
+        @NotNull String chromosomeEnd,
```
ArnoHMF: Didn't we change the names for DNA fusions from start/end to up/down? Wouldn't it make sense to keep these names consistent?

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaFusion.java
```diff
@@ -0,0 +1,56 @@
+package com.hartwig.hmftools.finding.datamodel;
+
+import com.hartwig.hmftools.finding.datamodel.finding.Finding;
+
+import org.jspecify.annotations.Nullable;
+
+import jakarta.validation.constraints.NotNull;
+
+@RecordBuilder
+public record RnaFusion(
```
ArnoHMF, line 10: Would it make sense to rename Fusion to DnaFusion? And would it not be RNAFusion/DNAFusion instead?

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/LowPurityTransformer.java
```diff
@@ -113,6 +113,11 @@ private static FindingRecord updateQCLowPurityStatus(@NotNull FindingRecord reco
                 || hasLowPurity(record.somaticDisruptions().status())
                 || hasLowPurity(record.fusions().status())
                 || hasLowPurity(record.viruses().status())
+                || hasLowPurity(record.rnaStatistics().status())
+                || hasLowPurity(record.highExpressionGenes().status())
+                || hasLowPurity(record.lowExpressionGenes().status())
+                || hasLowPurity(record.rnaFusions().status())
+                || hasLowPurity(record.novelSpliceJunctions().status())
```
ArnoHMF: More changes are required in this file to make this work correctly (in setLowPurityThresholds, updateLowPurityStatus)

ArnoHMF (commented):
Forgot to submit this. Looks good but had a few small comments.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaFusion.java
hongwingl: fixed

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/LowPurityTransformer.java
hongwingl: I asked Charles, RNA sample is separate and does not have a concept of low purity. Will remove these

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/Qc.java
```diff
@@ -10,8 +10,8 @@
 @RecordBuilder
 public record Qc(
         boolean isPass,
-        @NotNull SortedSet<QCStatus> warnings,
-        @NotNull SortedSet<QCStatus> errors,
+        @NotNull SortedSet<QcStatus> warnings,
+        @NotNull SortedSet<QcStatus> errors,
```
ArnoHMF, line 14: It's good to make this consistent, but then I personally think that QC follows the java convention better, assuming that this means QualityControl/QualityCheck?

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/util/LowPurityTransformer.java
ArnoHMF: Makes sense.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
```diff
@@ -1,26 +1,50 @@
...
+
     public enum Status
     {
         UNDETERMINED,
         HR_PROFICIENT,
         HR_DEFICIENT
     }
+
+    @RecordBuilder
+    public record Prediction(
```
ArnoHMF, line 46: Shouldn't this prediction have a likelihood?

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
```diff
@@ -1,26 +1,50 @@
...
+            @NotNull String findingKey,
+            @NotNull HrdCancerType cancerType,
```
ArnoHMF, line 48: This is an enum and for cuppa there is no enum. It would be nice if all cancer types returned are all according to the same ontology. Model wise there's nothing now that indicates that that is the case. I don't know enough about this though to suggest a solution

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/NovelSpliceJunction.java
```diff
@@ -0,0 +1,47 @@
...
+import com.hartwig.hmftools.finding.datamodel.finding.Finding;
+
+import jakarta.validation.constraints.NotNull;
+
+@RecordBuilder
+public record NovelSpliceJunction(
+        @NotNull String findingKey,
+        @NotNull String gene,
+        @NotNull String chromosome,
+        int junctionStart,
```
ArnoHMF, line 12: Shouldn't this also follow the up/down convention instead of start / end

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
```diff
@@ -1,26 +1,50 @@
...
+// vCHORD is cancer type specific (panels)
 @RecordBuilder
 public record HomologousRecombination(
-        @NotNull String findingKey,
-        @NotNull HomologousRecombination.Status status,
-        @NotNull ThresholdValue hrdValue,
-        double brca1Value,
-        double brca2Value,
-        @NotNull String hrdType,
+        @Nullable Collection<Prediction> predictions,
```
ArnoHMF: So the question I would have as a user of this model is if there are multiple values, which one is the one that I should use in the report? Any hints from the model would then be useful. Do I for instance take the top prediction (this is what we do for CUPPA)? And also for IVDR we would have to do some validation on these results and make some statements about what results can be used in what scenario.

ArnoHMF (commented):
One thing I was wondering about is if it would make sense to change the overal model a bit so we group the RNA and DNA results in separate records (and DNA could be split in somatic and germline or tumor and ref as well in principle). As it seems that both can be available independently and they both for instance have their own QC. Now it's not so clear from the model that the QC on finding record does not apply to the RNA results. Having them in two separate records with their own QC makes that much clearer. 

so you get something like record FindingRecord(String version, MetaProperties metaProperties, DNARecord dnaRecord, RNARecord rnaRecord) en then DNARecord of course contains most of current FindingRecord results. Or alternatively FindingRecord(String version, MetaProperties metaProperties, DNARecord tumorDNARecord, DNARecord refDNARecord, RNARecord rnaRecord), but that seems more complicated as there are some tumor only properties (so that has to be another type again).

Just thinking out loud here, and the first suggestion (split DNA/RNA) can of course be implemented independently of the second one (split tumor/ref)

Ideally the modeling gives guidance how the results can be used without the need for adding documentation.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/Qc.java
hongwingl: Interesting, you are right: https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms

> For 4+ letter acronyms, use mixed case. The standard library does this, and it just makes good sense.
> 
> For 3 letter acronyms, you can use all uppercase like the JCL, or you can use mixed case like the .NET Framework does. Either way, be consistent.
> 
> For 2 letter acronyms, use all uppercase.

I always thought should always be mixed case. Learn something everyday

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/NovelSpliceJunction.java
hongwingl: I asked Charles, he said splice junction should be start / end. I don't know it well enough so followed what he said

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
hongwingl: Will take the prediction with the correct cancer type. I could make it a map of HrdCancerType -> Prediction, that would make it clearer

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
hongwingl: This enum is pretty much for vCuppa. Cuppa use ALL, vCuppa would use the other ones.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/Qc.java
ArnoHMF: Ah. Wasn't aware that the rule is different for 4 letters. Agree that does make sense for 4 letters. But according to this DNA and RNA should also be uppercase

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/NovelSpliceJunction.java
ArnoHMF: Ok. Sounds fair ;-) Very possible that this has a slightly different meaning.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
ArnoHMF: Hmmm. In most cases I'm not a big fan of using a map in the json model, but I agree that that could be a bit clearer as that doesn't put one type over another. But still, not a fan of using a map. 

Are there actually multiple values possible per cancerType? From the current model it seems that there will be max one value per cancerType because you did not add likelihood

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/Qc.java
hongwingl: Actually newer style guides say camel case: https://google.github.io/styleguide/javaguide.html

Such that you do not get IOQC when it should be IoQc

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
hongwingl: hrdValue is the likelihood, it is 0 to 1

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/Qc.java
ArnoHMF: Ok. Them Qc it is. Learn something everyday :-)

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/HomologousRecombination.java
ArnoHMF: Ah ok. Makes sense.

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaFusion.java
```diff
@@ -0,0 +1,56 @@
...
+        @NotNull String findingKey,
+        @Nullable String geneUp,
```
ArnoHMF, line 12: The DNA version has transcripts as well. Is that not necessary here?

ArnoHMF (approved):
Looks good to me

finding-datamodel/src/main/java/com/hartwig/hmftools/finding/datamodel/RnaFusion.java
hongwingl: great question!

PR #806, opened 2026-06-03
Opr 324 clinical relevant gene copy number

LiekeSchoenmaker:


finding/src/main/java/com/hartwig/hmftools/finding/FindingConfig.java
```diff
@@ -48,4 +54,13 @@ public DriverGene getDriverGene(String gene)
     {
         return driverGenes.get(gene);
     }
+
+    @Nullable
+    public Boolean findClinicalRelevantGeneCopyNumber(String gene)
```
ArnoHMF: I would simplify this to boolean isClinicalRelevantGeneCopyNumber(String gene) and return false if there is no model. On the caller side there is no advantage to know that there was no model configured.

finding/src/main/java/com/hartwig/hmftools/finding/GainDeletionFactory.java
```diff
@@ -253,6 +257,57 @@ private static GainDeletion toGainDel(PurpleGainDeletion purpleGainDeletion,
...
+    private static GainDeletion ClinicalRelevantGeneCopyNumber(@NotNull PurpleGeneCopyNumber clinicalRelevantGeneCopyNumber,
+            ArmCopyNumberFactory cnPerChromosome, FindingConfig findingConfig, List<GainDeletion> gainDeletions)
+    {
+        Boolean presence = findingConfig.findClinicalRelevantGeneCopyNumber(clinicalRelevantGeneCopyNumber.gene());
+        if(presence != null && presence)
+        {
+            Set<String> reportableGainDeletionsGenes =
+                    gainDeletions.stream().map(GainDeletion::gene).collect(Collectors.toSet());
+            if(!reportableGainDeletionsGenes.contains(clinicalRelevantGeneCopyNumber.gene()))
+            {
```
ArnoHMF: This part can also be done in the filter earlier:

```
    public static List<GainDeletion> ClinicalRelevantGeneCopyNumber(@NotNull List<PurpleGeneCopyNumber> clinicalRelevantGeneCopyNumbers,
            ArmCopyNumberFactory cnPerChromosome, FindingConfig findingConfig, @NotNull List<GainDeletion> gainDeletions)
    {
        return clinicalRelevantGeneCopyNumbers.stream()
                .filter(o -> findingConfig.findClinicalRelevantGeneCopyNumber(o.gene()))
                .filter(o -> gainDeletions.stream().noneMatch(gd -> gd.gene().equals(o.gene())))
                .map(o -> ClinicalRelevantGeneCopyNumber(o, cnPerChromosome, findingConfig, gainDeletions)).toList();
    }
```

The filtering on not null is then also not necessary anymore here in that filter.

finding/src/main/java/com/hartwig/hmftools/finding/GainDeletionFactory.java
```diff
@@ -253,6 +257,57 @@ private static GainDeletion toGainDel(PurpleGainDeletion purpleGainDeletion,
                 .build();
     }
 
+    @NotNull
```
ArnoHMF: @NotNull annotations are redundant in this code as it has marked @NotNull as default, so you can remove all @NotNull annotations

ArnoHMF (changes requested):

finding/src/main/java/com/hartwig/hmftools/finding/GainDeletionFactory.java
LiekeSchoenmaker: only the return of the function or also the parameters?

ArnoHMF: That's for all @NotNull annotations in this file, parameters and return values.

ArnoHMF (approved):
All good if you return that last @NotNull annotation for the return value

PR #812, opened 2026-06-17
ESVEE: use SAGA breakends for unchained phased assemblies

reecejones-hartwig:
- For phased assemblies with 0 or 1 links, and a SAGA match.
- Use the SAGA breakends rather than aligning the assembly.
- This allows resolving SGLs to paired breakends, and addresses cases where the existing alignments/breakends are wrong due to homology or poor mappability.
- Use the insert sequence from SAGA rather than what's implied by the phased assembly. This allows resolving the full insert sequence, and provides a consistent representation.
- Adjust breakends for homology, as the existing indel breakend routine did.
- Add `SAGAINF` VCF info tag to breakends which are inferred to exist based on extending the insert sequence with SAGA.

### DIFF
--- a/esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
+++ b/esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
@@ -136,10 +136,12 @@ public class AlignData
     public String cigar() { return mCigar; }
     public String xaTag() { return mXaTag; }
     public String mdTag() { return mMdTag; }
+
     public Orientation orientation() { return mOrientation; }
     public boolean isForward() { return mOrientation.isForward(); }
     public boolean isReverse() { return mOrientation.isReverse(); }
     public boolean isSupplementary() { return SamRecordUtils.isFlagSet(mFlags, SUPPLEMENTARY_ALIGNMENT); }
+
     public int leftSoftClipLength() { return mSoftClipLeft; }
     public int rightSoftClipLength() { return mSoftClipRight; }
     public int alignedBases() { return mAlignedBases; }
@@ -175,9 +177,7 @@ public class AlignData
     }
 
     public void markDroppedOnRequery() { mDroppedOnRequery = true; }
-
     public boolean droppedOnRequery() { return mDroppedOnRequery; }
-
     public boolean isRequery() { return mIsRequery; }
 
     public void setRequeriedSequenceCoords(int sequenceStart, int sequenceEnd)
@@ -189,17 +189,14 @@ public class AlignData
     }
 
     public int sequenceStart() { return mSequenceStart; }
-
     public int sequenceEnd() { return mSequenceEnd; }
 
     public int rawSequenceStart() { return mRawSequenceStart; }
-
     public int rawSequenceEnd() { return mRawSequenceEnd; }
 
     public List<CigarElement> cigarElements() { return mCigarElements; }
 
     public int adjustedAlignment() { return mAdjustedAlignment; }
-
     public double modifiedMapQual() { return mModifiedMapQual; }
 
     private int calcAdjustedIndelScore()
@@ -324,11 +321,8 @@ public class AlignData
     }
 
     public boolean hasLowMapQualAlignment() { return mSelectedAltAlignment != null || mUnselectedAltAlignments != null; }
-
     public boolean hasSelectedAltAlignment() { return mSelectedAltAlignment != null; }
-
     public boolean hasLowMapQualShortSvLink() { return mHasLowMapQualShortSvLink; }
-
     public AlternativeAlignment selectedAltAlignment() { return mSelectedAltAlignment; }
 
     public void setSelectedAltAlignments(
--- a/esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
+++ b/esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
@@ -19,7 +19,6 @@ import static com.hartwig.hmftools.esvee.assembly.AssemblyConstants.BWA_GAP_OPEN
 import static com.hartwig.hmftools.esvee.assembly.AssemblyConstants.MULTI_MAPPED_ALT_ALIGNMENT_REGIONS_V37;
 import static com.hartwig.hmftools.esvee.assembly.AssemblyConstants.MULTI_MAPPED_ALT_ALIGNMENT_REGIONS_V38;
 import static com.hartwig.hmftools.esvee.assembly.types.RepeatInfo.calcTrimmedBaseLength;
-import static com.hartwig.hmftools.esvee.common.FilterType.MIN_LENGTH;
 import static com.hartwig.hmftools.esvee.common.SvConstants.MIN_VARIANT_LENGTH;
 
 import static htsjdk.samtools.CigarOperator.D;
@@ -39,6 +38,7 @@ import com.hartwig.hmftools.common.genome.refgenome.RefGenomeVersion;
 import com.hartwig.hmftools.common.genome.region.Orientation;
 import com.hartwig.hmftools.common.region.ChrBaseRegion;
 import com.hartwig.hmftools.esvee.assembly.types.RepeatInfo;
+import com.hartwig.hmftools.esvee.common.saga.SagaAlignment;
 
 import org.broadinstitute.hellbender.utils.bwa.BwaMemAlignment;
 
@@ -46,7 +46,7 @@ import htsjdk.samtools.CigarElement;
 
 public class AlignData
 {
-    private ChrBaseRegion mRefLocation;
+    private final ChrBaseRegion mRefLocation;
     private final int mMapQual;
     private final int mNMatches;
     private final int mScore;
@@ -59,11 +59,11 @@ public class AlignData
     private int mSoftClipLeft;
     private int mSoftClipRight;
 
-    private Orientation mOrientation;
+    private final Orientation mOrientation;
     private final int mAlignedBases;
 
-    private int mRawSequenceStart;
-    private int mRawSequenceEnd;
+    private final int mRawSequenceStart;
+    private final int mRawSequenceEnd;
     private int mSequenceStart;
     private int mSequenceEnd;
     private boolean mDroppedOnRequery;
@@ -109,7 +109,7 @@ public class AlignData
         }
 
         mOrientation = SamRecordUtils.isFlagSet(mFlags, READ_REVERSE_STRAND) ? REVERSE : FORWARD;
-        mAlignedBases = mCigarElements.stream().filter(x -> x.getOperator() == M).mapToInt(x -> x.getLength()).sum();
+        mAlignedBases = mCigarElements.stream().filter(x -> x.getOperator() == M).mapToInt(CigarElement::getLength).sum();
 
         mSequenceStart = sequenceStart;
         mSequenceEnd = max(sequenceEnd - 1, 0);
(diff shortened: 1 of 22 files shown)


hmf-common/src/main/java/com/hartwig/hmftools/common/bam/CigarUtils.java
```diff
@@ -248,6 +247,12 @@ public static int getReadIndexFromPosition(final int alignmentStart, final List<
 
     public static int getReadIndexFromPosition(
             final int alignmentStart, final List<CigarElement> cigarElements, int position, boolean lastIfGapped, boolean inferFromSoftClip)
+    {
+        return getReadIndexFromPosition(alignmentStart, cigarElements, position, lastIfGapped ? -1 : 0, inferFromSoftClip);
+    }
+
+    public static int getReadIndexFromPosition(
```
reecejones-hartwig, line 254: Added the ability to choose between the previous or next index if the position lands in a gap

esvee/src/main/java/com/hartwig/hmftools/esvee/common/saga/SagaAssembly.java
```diff
@@ -9,42 +9,61 @@ public record SagaAssembly(
         SagaVariant variant,
         // Junctions occur just before each of these indices in the assembly sequence.
         // Usually length 2. Can be length 1 for DELs where the junction is a single position.
+        // E.g.
+        // sequence = RRRJJJRRR
+        // junctionOffsets[0] = 3
+        // junctionOffsets[1] = 6
         List<Integer> junctionOffsets,
-        int assemblyLength
+        String sequence
```
reecejones-hartwig, line 17: The SAGA fasta is total ~85MB, so it should be fine to store the sequences, from a memory usage perspective

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
```diff
@@ -71,6 +79,101 @@ public void formBreakends(final List<AlignData> alignments)
...
+            indelLength = rawInsertSequence.length();
+
+            lowerBreakendInferred = startInferredLength > 0;
+            upperBreakendInferred = endInferredLength > 0;
+        }
+        else
+        {
+            // Deletion - no insert sequence and the assembly spans the junction (otherwise it couldn't have matched the SAGA variant).
+
+            // FIXME: the assembly can have a few different number of bases deleted and that is lost here, since in the saga alignment small indels are allowed around the junction
```
reecejones-hartwig: When aligning the phased assembly to SAGA, we allow small (<= 2b) indels near the junctions. This means that the phased assembly could imply a deletion of +/- a few bases, which is effectively lost when we use the SAGA breakends here.
This could be fixed if we care that much, but I don't think it's a trivial fix.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
```diff
@@ -131,76 +272,75 @@ private boolean formIndelBreakends(final AlignData alignment)
...
+            @Nullable
+            HomologyData homology,
+            int assemblyIndelStart,
+            int assemblyIndelEnd
+    )
+    {
+    }
 
-        int indelSeqEnd = indelSeqStart + (indelCoords.isInsert() ? indelCoords.Length : 1);
+    private IndelBreakendData calcIndelBreakendsWithHomology(final String chromosome, final IndelCoords indelCoords, int assemblyIndelStart,
```
reecejones-hartwig, line 291: Potential remaining TODO for a later commit: adjust this to allow inexact homology, for the case of large SAGA inserts.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignmentFragments.java
```diff
@@ -611,10 +611,10 @@ private boolean readSpansJunction(final Breakend breakend, final SupportRead rea
             {
                 int[] indelSequenceIndices = segment.indelSeqenceIndices();
 
-                if(readSeqIndexStart < indelSequenceIndices[0] && readSeqIndexEnd > indelSequenceIndices[0])
+                if(readSeqIndexStart < indelSequenceIndices[0] && readSeqIndexEnd >= indelSequenceIndices[0])
```
reecejones-hartwig, line 614: There was an off by one error here, where a read starting exactly on indelSequenceIndices[0] wouldn't be counted as crossing the junction. That is now fixed and the maths checks out in all cases.

For insertion:
- indelSequenceIndices[0] is the index of the first inserted base
- indelSequenceIndices[1] is the index of the first ref base after the insert sequence

For deletion:
- indelSequenceIndices[0] is the index of the first ref base after the deleted bases
- indelSequenceIndices[1] is the same

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
```diff
@@ -370,6 +394,15 @@ public static AlignData from(final BwaMemAlignment alignment, final RefGenomeVer
                 alignment.getSamFlag(), alignment.getCigar(), alignment.getNMismatches(), alignment.getXATag(), alignment.getMDTag());
     }
 
+    public static AlignData fromSaga(final SagaAlignment alignment)
+    {
+        BwaMemAlignment rawAlignment = alignment.rawAlignment();
+        return new AlignData(
+                new ChrBaseRegion(alignment.sagaVariant().id(), alignment.sagaStart() + 1, alignment.sagaEnd()),
```
reecejones-hartwig, line 377: I think this is only used for informative output, so we can mostly fill in whatever information. Here I'm pretending the SAGA variant assembly is another ref contig to which the alignment refers.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
```diff
@@ -126,25 +126,41 @@ public AlignData(
     }
 
     public ChrBaseRegion refLocation() { return mRefLocation; }
+
```
reecejones-hartwig: IntelliJ code formatting

charlesshale: I'm not a fan of Intel-J adding lines between simple getter/setter methods. Can we disable that? also adds more changes unnecessarily to the PR. Can discuss at the meeting tomorrow.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AssemblyAligner.java
```diff
@@ -126,15 +127,62 @@ public void processAssembly(final AssemblyAlignment assemblyAlignment)
...
+    @Nullable
+    private AlignmentInfo processSagaMatchedAssembly(final AssemblyAlignment assemblyAlignment)
+    {
+        SagaMatchBySequence sagaMatch = assemblyAlignment.sagaMatch();
+        if(sagaMatch == null)
+        {
+            return null;
+        }
+
+        // FIXME? validate that the saga match is nearby to the assembly.
```
charlesshale, line 185: Is this important? can discuss in Esvee chat

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
```diff
@@ -71,6 +79,101 @@ public void formBreakends(final List<AlignData> alignments)
...
+            return false;
+        }
+
+        SagaAssembly sagaAssembly = sagaAlignment.sagaAssembly();
+        SagaVariant sagaVariant = sagaAssembly.variant();
+        SagaBreakend lowerSagaBreakend = sagaVariant.breakend1();
+        SagaBreakend upperSagaBreakend = sagaVariant.breakend2();
+
+        // SAGA variants should only be indels, but check because other types are not handled here.
+        if(!lowerSagaBreakend.chromosome().equals(upperSagaBreakend.chromosome()))
```
charlesshale, line 99: Is a more thorough check required here? is it possible to get something other than an INDEL if the SAGA resource has none?

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AlignData.java
reecejones-hartwig: I've adjusted the code style file so IntelliJ won't change this, and will revert the diff here.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AssemblyAligner.java
reecejones-hartwig: In my analysis I found about 250 phased assemblies where the SAGA match is not within 10kb of any of the junctions. I looked into some of them and couldn't discern any actionable fixes.

Potential options:

- Continue to investigate those 250 cases and try to fix the root causes.
- Add a check in ESVEE to discard the SAGA match if it's too far away from the junctions.
- Do nothing because it's too few cases.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
reecejones-hartwig: To be clear, the SAGA resource only has indels. There will never be an indel in practice. However the SAGA code does not strictly forbid a different type of variant because they are simply stored as two breakends with independent chromosome and orientation. So I wanted to be defensive in checking this in breakend building, because it will silently fail if the data is not as expected.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
```diff
@@ -71,6 +78,99 @@ public void formBreakends(final List<AlignData> alignments)
...
     }
 
+    public boolean formBreakendsFromSaga(final SagaAlignment sagaAlignment, final AlignData alignData)
+    {
+        // For a SAGA-matched variant, use the breakends stated in the SAGA resource. Why?
+        //   - Some cases where homology causes alignment to be misleading, resulting in wrong breakends.
+        //   - Recover SGLs to full variants given that one half matches a SAGA variant.
+
+        // Only use SAGA breakends for the simple case of 0 or 1 links. Chained assemblies require more care, not implemented for now.
+        boolean isChained = mAssemblyAlignment.phaseSet() != null && mAssemblyAlignment.phaseSet().assemblyLinks().size() > 1;
```
charlesshale: Move this check right to the start of alignment - ie AssemblyAligner - and so don't call into SAGA for these

esvee/src/main/java/com/hartwig/hmftools/esvee/common/saga/SagaSequenceMatcher.java
```diff
@@ -86,7 +81,7 @@ private SagaMatchBySequence matchFromAlignments(final MatchArguments args, final
         return candidates.stream()
                 .filter(SagaSequenceMatchCandidate::isAccepted)
```
charlesshale, line 82: If you'd like to keep the filters (eg "reverse_strand") make enums for them
Also fine to just mark a candidate as filters T/F

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AssemblyAligner.java
```diff
@@ -126,15 +127,62 @@ public void processAssembly(final AssemblyAlignment assemblyAlignment)
         if(mSagaMatcher != null)
         {
             List<SagaJunctionInfo> junctionInfos = assemblyAlignment.linkIndices().stream().map(SagaJunctionInfo::new).toList();
-            boolean lowerJunctionOverlap = assemblyAlignment.assemblies().size() == 1 && assemblyAlignment.assemblies().get(0).hasLineSequence();
+            boolean lowerJunctionOverlap = assemblyAlignment.assemblies().stream().anyMatch(JunctionAssembly::hasLineSequence);
```
charlesshale: Could you add a comment to explain the relevance of isLine T/F for latter junction overlap conditions. ie the link back to LINE vs non-LINE extension length constants. Maybe make a note to alternatively work this out the min overlap lengths vs assembly extension length later on.

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/BreakendBuilder.java
charlesshale: I would also be fine to remove this check and assume that the SAGA resources is constructed only from INDELs. And then revisit this if we ever then support more complex types.

charlesshale (approved):

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/AssemblyAligner.java
reecejones-hartwig: Continue discussion in chat

PR #813, opened 2026-06-18
INFRA-2915 - Add analysis_date config param, make sampling_date optional

rhassaine:
## Initial Issue
The optional sampling_date flag in ORANGE was a potentially confusing label and the underlying logic of that flag was to default to local date when not provided. This resulted in some users assuming sampling_date and sample biopsy dates to be identical when in fact not. 

## Changes
- `analysis_date` is a new optional config param that defaults to the current date if not provided
- `sampling_date` is now optional and defaults to `null` if not provided (previously defaulted to current date)
- Updated `OrangeRecord` datamodel to reflect both fields
- PDF report now always shows analysis_date under the "Date Analysed", sampling_date appears as an additional row only when provided
- Refactored interpretSamplingDateParam & interpretAnalysisDateParam into a single generic interpretDateParam(configBuilder, argName, defaultValue)
- Fixed ReportGeneratorTestApplication to write output to ~/hmf/temp/ correctly & auto-crate the directory if absent
- Added `OrangeConfigTest` to verify the new behavior & analysisDate to orange-datamodel test JSON fixtures

## Testing
6 new unit tests added covering:
- Valid date parsing for both fields
- `sampling_date` returns null when not provided or invalid
- `analysis_date` defaults to today when not provided or invalid


orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -74,6 +74,8 @@
 import com.hartwig.hmftools.datamodel.orange.OrangeRefGenomeVersion;
 import com.hartwig.hmftools.orange.util.PathResolver;
 
+import javax.swing.*;
```
DFKoetsier: This isn't useful, right?

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
```diff
@@ -340,11 +354,28 @@ private static LocalDate interpretSamplingDateParam(final String samplingDateStr
         }
         catch(DateTimeParseException exception)
         {
-            samplingDate = LocalDate.now();
+            samplingDate = null;
             LOGGER.warn("Could not parse configured sampling date '{}'. Expected format is '{}'", samplingDateString, format);
         }
         return samplingDate;
     }
+    static LocalDate interpretAnalysisDateParam(final String analysisDateString)
```
DFKoetsier: A bit redundant with `interpretSampleDateParam`. I think it would be better to make a more generic `interpretDateParam` method that takes configBuilder, the command line argument and the default value as arguments. Then you could also include the "if config doesn't have the argument" part.

orange/src/test/java/com/hartwig/hmftools/orange/TestOrangeConfigFactory.java
```diff
@@ -39,17 +39,17 @@ public static OrangeConfig createMinimalConfig()
...
-                LILAC_DIR, CHORD_DIR, CUPPA_DIR, PEACH_DIR,SIGS_DIR, VIRUS_DIR, ISOFOX_DIR, false);
+                LILAC_DIR, CHORD_DIR, CUPPA_DIR, PEACH_DIR, SIGS_DIR, VIRUS_DIR, ISOFOX_DIR, false);
     }
 
     public static OrangeConfig createTargetedConfig()
     {
         return new OrangeConfig(
                 ExperimentType.TARGETED, TUMOR_SAMPLE_ID, null, null,
-                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),
+                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),LocalDate.now(),
```
DFKoetsier: Missing space

orange/src/test/java/com/hartwig/hmftools/orange/TestOrangeConfigFactory.java
```diff
@@ -59,7 +59,7 @@ public static OrangeConfig createWGSConfigTumorOnly()
     {
         return new OrangeConfig(
                 ExperimentType.WHOLE_GENOME, TUMOR_SAMPLE_ID, null, null,
-                RefGenomeVersion.V37, Collections.emptySet(), LocalDate.now(),
+                RefGenomeVersion.V37, Collections.emptySet(), LocalDate.now(),LocalDate.now(),
```
DFKoetsier: Missing space

orange/src/test/java/com/hartwig/hmftools/orange/TestOrangeConfigFactory.java
```diff
@@ -69,7 +69,7 @@ public static OrangeConfig createWGSConfigTumorNormal()
     {
         return new OrangeConfig(
                 ExperimentType.WHOLE_GENOME, TUMOR_SAMPLE_ID, REFERENCE_SAMPLE_ID, null,
-                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),
+                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),LocalDate.now(),
```
DFKoetsier: Missing space

orange/src/test/java/com/hartwig/hmftools/orange/TestOrangeConfigFactory.java
```diff
@@ -79,7 +79,7 @@ public static OrangeConfig createWGTSConfigTumorNormal()
     {
         return new OrangeConfig(
                 ExperimentType.WHOLE_GENOME, TUMOR_SAMPLE_ID, REFERENCE_SAMPLE_ID, "tumor_sample",
-                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),
+                RefGenomeVersion.V37, Set.of(MELANOMA_DOID), LocalDate.now(),LocalDate.now(),
```
DFKoetsier: Missing space

orange/README.md
```diff
@@ -40,16 +40,17 @@ java -jar orange.jar \
...
+| primary_tumor_location   | Printed at top of report                                                                                                      | 
+| experiment_type          | WGS or PANEL                                                                                                                  |
+| rna_sample_id            | Used to display RNA sample genotype values from SageAppend                                                                    |
+| panel_name               | Optional, for display only                                                                                                    
+| add_disclaimer           | If set, adds a "research use only" disclaimer to the footer of every page.                                                    |  
+| pipeline_sample_root_dir | Optional, all individual algo paths are derived from this path, assuming the pipeline has been run using HMF pipeline         |
+| sample_data_dir          | Optional, all data is expected to exist in the root of this path                                                              |
```
DFKoetsier: Table isn't formatted correctly. Right bars aren't at the same position

DFKoetsier (changes requested):
Mostly looks good.

Larger problems:
Not all tests pass when building. Orange datamodel tests don't all work.

Making a test report also crashes. The sampling date is printed in FrontPageTables under "Date Analysed". This should probably be changed to the analysis date.

I also have a few minor comments.

orange/src/main/java/com/hartwig/hmftools/orange/OrangeConfig.java
rhassaine: Yes, implemented

DFKoetsier (changes requested):
Looks good, except for the new .gitignore merge conflict, and the fact findings has failing tests. You need to at least fix `metaPropertiesBuilder()` in `TestFindingRecordFactory.java`. Maybe more is needed.

finding-datamodel/findings.json
DFKoetsier, line 1: I don't think you want to commit this file. I think `testWriteRead` in `FindingsJsonTest.java` creates this file, but it shouldn't. 

Maybe change the test to use `Files.createTempFile` instead, or add it to the gitignore, or just uncommit the file and let someone else fix the test. I would just uncommit it.

DFKoetsier (approved):
One tiny change left. After that, LGTM!

finding-datamodel/findings.json
rhassaine: Went for your suggestion to use `Files.createTempFile`

PR #824, opened 2026-06-29
Esvee: indel inexact homology (AUS-260)

reecejones-hartwig:
In indel breakend homology checking, allow SNVs using BWA-like scoring. This affects both DEL breakend adjustment and INS to DUP breakend adjustment.

For now, only handle SNVs. A future improvement is to handle indels in the homology sequence using a more complete alignment routine.


esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/HomologyData.java
```diff
@@ -188,35 +196,58 @@ public static HomologyData determineIndelHomology(final String refBasesStart, fi
...
+        if(exactMatch == 0)
+            return NO_HOMOLOGY;
+        int exactStart = (exactMatch + 1) / 2; // round up if an odd length
+        int exactEnd = exactMatch - exactStart;
 
-        for(int i = 0; i < overlap; ++i)
+        // Continue on to find inexact homology.
+        // Emulate BWA-MEM scoring, to find the point at which the alignment would clip.
+        // This makes it consistent with the non-indel code which uses split alignments to determine homology.
+        // FIXME? handle indels?
```
charlesshale: Is this is a NOTE? or do you intend to add support for indels?

I am fine to just mention they aren't handled

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/HomologyData.java
```diff
@@ -188,35 +196,58 @@ public static HomologyData determineIndelHomology(final String refBasesStart, fi
...
+        {
+            ++exactMatch;
+        }
+        if(exactMatch == 0)
+            return NO_HOMOLOGY;
+        int exactStart = (exactMatch + 1) / 2; // round up if an odd length
+        int exactEnd = exactMatch - exactStart;
 
-        for(int i = 0; i < overlap; ++i)
+        // Continue on to find inexact homology.
```
charlesshale: Could you keep to the commenting convention I've used elsewhere in Esvee - ie lowercase, no full-stops if one-liners. See how other comments look.

charlesshale (approved):

esvee/src/main/java/com/hartwig/hmftools/esvee/assembly/alignment/HomologyData.java
reecejones-hartwig: I may add support for indels if I get time and we find evidence to suggest it needs to be done. Otherwise I would like to move on to other work.

PR #833, opened 2026-07-02
INFRA-2908 bamtools docker image was missing sambamba which is needed…

mvanniekerkHartwig:
… for bamchecker


HMFstefan (commented):
Looks good!

@charlesshale could you please confirm the version of Sambamba is suitable for bamChecker?

charlesshale (approved):

PR #840, opened 2026-07-08
INFRA-2980 fix inconsistent behaviour compared to redux mate referenc…

mvanniekerkHartwig:
…e * with a non-zero mate alignment start


mvanniekerkHartwig:
Fixed in https://github.com/hartwigmedical/hmftools/releases/tag/bam-tools-v1.6.2-beta.10

PR #841, opened 2026-07-08
FastqTools: fix truncation data loss and match-count race

kzuberihmf:
PR for a couple of fixes to merge back into the AUS440-fastq-misc branch.

* There's a missing writer.close(), the last batch of data sits in the buffer and isn't written. Fixed properly using try with resources.
* A count is incremented from threads which is non deterministic, fixed with atomic. Don't think this should introduce much contention.

Both issues reproduced using machine generated test scripts + data, and the fixes verified same way.


pauldwolfe (approved):

kzuberihmf:
Will merge to the other branch and try tagging a fastq-tools-v1.1-beta.2.

charlesshale:
thanks guys. I'll review the code in the branch now that it's been merged.

PR #842, opened 2026-07-10
OPR-469: Support cuppa chart plot to findings model

LiekeSchoenmaker:
Support the cuppa chart plot in the output of the findings model to use for OncoAct reporting

### DIFF
--- a/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangePlots.java
+++ b/orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangePlots.java
@@ -49,7 +49,6 @@ public interface OrangePlots
     @Nullable
     String cuppaFeaturePlot();
 
-    @Deprecated
     @Nullable
     String cuppaChartPlot();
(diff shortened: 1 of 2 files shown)


orange-datamodel/src/main/java/com/hartwig/hmftools/datamodel/orange/OrangePlots.java
```diff
@@ -49,7 +49,6 @@ public interface OrangePlots
     @Nullable
     String cuppaFeaturePlot();
 
-    @Deprecated
```
ArnoHMF, line 52: No need to remove this. If Sydney says this is deprecated it probably is. It won't affect the working anyway.

ArnoHMF (approved):
This looks good to me. 

However, it seems that there is more work to be done here in the future. Because this plot seems deprecated. Either we need to change the plot in the OncoAct report, or we should change the findings datamodel so it supports multiple plots

LiekeSchoenmaker:
Solution will be implemented in OncoAct repo,

PR #843, opened 2026-07-10
Infra 2908 - add bamchecker option to filter out reads+contigs not in the ref genome

mvanniekerkHartwig:
Bamchecker uses the provided reference enough to read/process the BAM and repair pipeline-relevant issues, but it does not normalize the BAM to that reference’s sequence dictionary. It preserves the original BAM header contigs and can preserve reads aligned to contigs that are absent from the supplied HMF FASTA.
Most of the normal pipeline tolerates that because it mostly queries/works on the human contigs it cares about. So the BAM can be “good enough” for OA/pipeline processing even if its header still says “this BAM also contains CMV/HBV/HCV/HPV/etc.”
CRAM conversion is stricter. When samtools writes CRAM with -T Homo_sapiens_assembly38.alt.masked.fasta, it needs the reference sequence for every contig it is asked to encode in the CRAM header/records. If the BAM says https://github.com/sq SN:CMV or contains reads on HCV-1, but the FASTA has no CMV or HCV-1, samtools cannot build a normal externally referenced CRAM.

Note that I'm not completely sure whether this approach is the right one. I'm also not sure whether the implementation makes sense, because I'm not too familiar with bam-tools.

### DIFF
--- a/bam-tools/src/main/java/com/hartwig/hmftools/bamtools/checker/BamChecker.java
+++ b/bam-tools/src/main/java/com/hartwig/hmftools/bamtools/checker/BamChecker.java
@@ -61,9 +61,17 @@ public class BamChecker
                 .referenceSequence(new File(mConfig.RefGenomeFile))
                 .open(new File(mConfig.BamFile)).getFileHeader();
 
-        FragmentCache fragmentCache = new FragmentCache(fileHeader);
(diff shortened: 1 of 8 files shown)


mvanniekerkHartwig:
Fixed in https://github.com/hartwigmedical/hmftools/releases/tag/bam-tools-v1.6.2-beta.10

PR #852, opened 2026-07-16
Fix for SBX YC tag parsing when encountering empty duplex region

rhassaine:
Fixes a crash in `getDuplexIndelIndices` when the YC tag's duplex segment is empty (e.g. a fully-simplex read with no duplex coverage). `String`.split() without a limit silently drops trailing empty tokens, so "390+".split("\+") returns ["390"] instead of ["390", ""], failing the length check and returning null, which is now fatal. 

**Fix**: pass -1 as the limit to preserve the empty trailing token.
This tag shape is valid per the SBX-D YC tag spec, so 390++ should parse successfully rather than crash.

**Tested**: patched jar run against a real SBX CRAM containing this exact case and REDUX completes cleanly (379,999 reads processed, no invalid SBX ycTag error), vs. the unpatched jar which failed with invalid SBX ycTag(390++).

### DIFF
--- a/hmf-common/src/test/java/com/hartwig/hmftools/common/sequencing/SbxBamUtilsTest.java
+++ b/hmf-common/src/test/java/com/hartwig/hmftools/common/sequencing/SbxBamUtilsTest.java
@@ -16,18 +16,18 @@ public class SbxBamUtilsTest
     public void
     testGetDuplexIndelsNoDuplexIndels()
     {
-        String ycTagStr = "0+100+0";
+        String ycTagStr = "+100+"; // no leading simplex count or trailing tail, both meaning zero
         List<Integer> duplexIndelIndices = getDuplexIndelIndices(ycTagStr);
 
         assertTrue(duplexIndelIndices.isEmpty());
 
-        ycTagStr = "10+100I100+0";
+        ycTagStr = "10+100I100+"; // no trailing tail, meaning zero
         duplexIndelIndices = getDuplexIndelIndices(ycTagStr);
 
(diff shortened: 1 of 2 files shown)


hmf-common/src/test/java/com/hartwig/hmftools/common/sequencing/SbxBamUtilsTest.java
```diff
@@ -39,5 +39,11 @@ public class SbxBamUtilsTest
 
         assertEquals(1, duplexIndelIndices.size());
         assertTrue(duplexIndelIndices.contains(41));
+
+        ycTagStr = "390++"; // empty duplex region and omitted tail, not invalid
+
+        duplexIndelIndices = getDuplexIndelIndices(ycTagStr);
+
+        assertTrue(duplexIndelIndices.isEmpty());
```
DFKoetsier, line 47: Maybe also change the other test cases to remove initial and trailing zeros where they currently exist, and also add a case for `ycTagStr = "++"`. This should all already work with the current code, but it would be good to make sure the test cases fully match the new format.

DFKoetsier (approved):
Looks good! The fix is smaller than I was expecting.

I think it would be nice to further improve the test a bit, but otherwise LGTM

hmf-common/src/test/java/com/hartwig/hmftools/common/sequencing/SbxBamUtilsTest.java
rhassaine: Done

PR #855, opened 2026-07-20
Lilac: Fix somatic splice indel allele assignment (WAMS-37)

DFKoetsier:


DFKoetsier:
Fixed differently in this commit: https://github.com/hartwigmedical/hmftools/commit/67a2e9f157e0de70a6100be379e524a44bf3ae80

PR #863, opened 2026-07-28
Redux: skip invalid reads without truncating regions (AUS418)

shiv-hartwig:
Stops a single upstream-invalid read from silently truncating a REDUX region or aborting the run.

1. SILENT stringency - htsjdk returns odd records instead of throwing.
2. Skip and log malformed reads (zero-length CIGAR element or SEQ/CIGAR length mismatch), which otherwise abort BAM encode.
3. Log and continue on a read error rather than dropping the rest of the region.


shiv-hartwig:
not required, change already in the latest REDUX release

PR #867, opened 2026-08-05
PanelBuilder: use hmf-common BwaMemAligner (AUS-471)

reecejones-hartwig:
Switch to new hmf-common BwaMemAligner. Identical behaviour, validated identical output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


panel-builder/src/main/java/com/hartwig/hmftools/panelbuilder/probequality/Utils.java
reecejones-hartwig, line 1: Functionality already exists in hmf-common. No longer needed here

charlesshale (approved):

PR #868, opened 2026-08-06
PanelBuilder: exon-aware RNA transcript panel support (AUS-422)

reecejones-hartwig:
Adds exon-aware RNA transcript panel support.

- New `-genes_rna` input: produces probes over the exonic (transcribed) sequence of selected genes, written to separate `rna_*` output files.
  - Coding sequence always covered; 5' and 3' UTR optional per gene.
  - Canonical transcript by default, or a specified transcript subset.
  - Probes stay within exon sequence and splice across a junction only to pad a short exon, improving capture of both known transcripts and novel splices/fusions.
- Renamed existing DNA output files to start with `dna_`
- Fixes the quality score of non-contiguous probes (structural-variant and RNA spliced probes), which previously overstated off-target risk.
- Probe output extra info now includes the transcript id (for both DNA and RNA).
- Tolerates Excel-added double quotes around the transcript-names input field.

Validation:
- Unit tests
- Manual inspection in IGV by Peter and I
- Diff to previous PanelBuilder output


panel-builder/src/main/java/com/hartwig/hmftools/panelbuilder/samplevariants/GermlineMutation.java
reecejones-hartwig, line 1: I reformatted all PanelBuilder code in line with latest Java code style

charlesshale (approved):

